PR #2, opened 2021-01-19
Explore how to prove first implementation obligation for fake/trivial implementation

mark-moir:
Note: This pull request is really to enable discussion and interaction; maybe we'll decide to merge it or go in a different direction.

We explore how to prove ` VO.ImplObligation₁` using `unwind` and `Any-Step-elim`, aided by a new `Any-Step-⇒` that allows us to establish properties of the first transition to send a given signature.

We have made some progress, which has helped to highlight some challenges to discuss, and also to give a sense of what's more painful than it should be, so we can think about how to streamline before moving on to harder properties and/or a more realistic implementation.

A few postulates have been used and not yet proved.  Most are about the implementation handler and are hopefully straightforward.  However, I struggled trying to prove`epochPreservestoNodeId`, which seems like it should be straightforward; would appreciate input from Agda experts).  Apropos of nothing, hi @VictorCMiraldo :-).

### DIFF
--- a/LibraBFT/Abstract/Properties.agda
+++ b/LibraBFT/Abstract/Properties.agda
@@ -52,11 +52,6 @@ module LibraBFT.Abstract.Properties
        → vMember v' ≡ α → HasBeenSent v'
        → vRound v ≡ vRound v'
        → vBlockUID v ≡ vBlockUID v'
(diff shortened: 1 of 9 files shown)


mark-moir:
Superseded by #18.   Closing.

PR #9, opened 2021-02-03
Proved bft-lemma for one peer, one vote and completed IsValidQC

lisandrasilva:
Proved the well-known result that for `N` peers of which at most `f` are byzantine, for any two lists of at least `N - f` different peers, there is at least one honest peer in their intersection.
This can be a utility for someone trying to create an abstract `EpochConfig` for an implementation that assumes one peer, one vote. 

Also completed the hole in `IsValidQC` using `All-reduce` function.

### DIFF
--- a/LibraBFT/Abstract/BFT.agda
+++ b/LibraBFT/Abstract/BFT.agda
@@ -6,11 +6,479 @@
 open import LibraBFT.Prelude
 open import LibraBFT.Lemmas
 open import LibraBFT.Abstract.Types
+open import LibraBFT.Base.PKCS
 
-module LibraBFT.Abstract.BFT where
+
+module LibraBFT.Abstract.BFT
+  (authorsN  : ℕ)
+  (bizF      : ℕ)
+  (isBFT     : authorsN ≥ suc (3 * bizF))
+  (getPubKey : Fin authorsN → PK)
+
+
+
+ where
 
   -- This is a utility function to make it easy to provide the bft-assumption
   -- for the abstract EpochConfig by by assuming that at most bizF members are byzantine
   -- and that authorsN ≥ suc (3 * bizF) and that a list of Members is a quorum if it
   -- contains at least authorsN ∸ bizF distinct Members.
+
+ QSize : ℕ
+ QSize = authorsN ∸ bizF
+
+ -- The set of members of this epoch.
+ Member : Set
+ Member = Fin authorsN
+
+ Meta-dishonest? :  ∀ (m : Member) → Dec (Meta-Dishonest-PK (getPubKey m))
+ Meta-dishonest? m = Meta-DishonestPK? (getPubKey m)
+
+ module _  (bft-assumption : ∀ {xs : List Member} → IsSorted _<Fin_ xs
+                             → length (List-filter Meta-dishonest? xs) ≤ bizF)
+   where
+
+   intersectElem : List Member → Member → List Member
+   intersectElem [] y = []
+   intersectElem (x ∷ xs) y
+      with Fin-<-cmp x y
+   ... | tri< a ¬b ¬c = intersectElem xs y
+   ... | tri≈ ¬a b ¬c = y ∷ []
+   ... | tri> ¬a ¬b c = []
+
+   _∈?_ : (x : Member) → (xs : List Member) → Dec (Any (x ≡_) xs)
+   x ∈? xs = Any-any (x ≟Fin_) xs
+
+   intersect : List Member → List Member → List Member
+   intersect xs [] = []
+   intersect xs (y ∷ ys) = intersectElem xs y ++ intersect xs ys
+
+
+
+   unionElem : List Member → Member → List Member
+   unionElem [] y = y ∷ []
+   unionElem (x ∷ xs) y
+     with Fin-<-cmp x y
+   ... | tri< a ¬b ¬c = x ∷ unionElem xs y
+   ... | tri≈ ¬a b ¬c = x ∷ xs
+   ... | tri> ¬a ¬b c = y ∷ x ∷ xs
+
+   union : List Member → List Member → List Member
+   union xs [] = xs
+   union xs (y ∷ ys) = unionElem (union xs ys) y
+
+
+   intDiffElem : ∀ {slist} (xs : List Member)
+                 → (intS : IsSorted _<Fin_ slist)
+                 → (y : Member)
+                 → OnHead _<Fin_ y slist
+                 → IsSorted _<Fin_ (intersectElem xs y ++ slist)
+   intDiffElem [] slist y x = slist
+   intDiffElem (x ∷ xs) slist y yOH
+     with Fin-<-cmp x y
+   ... | tri< a ¬b ¬c = intDiffElem xs slist y yOH
+   ... | tri≈ ¬a b ¬c = yOH ∷ slist
+   ... | tri> ¬a ¬b c = slist
+
+
+   trans-OnHead : ∀ {xs : List Member} {y x : Member}
+          → OnHead _<Fin_ y xs
+          → x <Fin y
+          → OnHead _<Fin_ x xs
+   trans-OnHead [] x<y = []
+   trans-OnHead (on-∷ y<f) x<y = on-∷ (Fin-<-trans x<y y<f)
+
+   ++-OnHead : ∀ {xs ys : List Member} {y : Member}
+                      → OnHead _<Fin_ y xs
+                      → OnHead _<Fin_ y ys
+                      → OnHead _<Fin_ y (xs ++ ys)
+   ++-OnHead {[]} {ys} {y} xsOH ysOH = ysOH
+   ++-OnHead {x ∷ xs} {ys} {y} (on-∷ y<x) ysOH = on-∷ y<x
+
+   intElem-OnHead : ∀ {xs : List Member} {y x : Member}
+                    → x <Fin y
+                    → OnHead _<Fin_ x (intersectElem xs y)
+   intElem-OnHead {[]} {y} {x} x<y = []
+   intElem-OnHead {x₁ ∷ xs} {y} {x} x<y
+      with Fin-<-cmp x₁ y
+   ... | tri< a ¬b ¬c = intElem-OnHead {xs} x<y
+   ... | tri≈ ¬a b ¬c = on-∷ x<y
+   ... | tri> ¬a ¬b c = []
+
+   int-OnHead : ∀ {xs ys : List Member} {y : Member}
+         → IsSorted _<Fin_ (y ∷ ys)
+         → OnHead _<Fin_ y (intersect xs ys)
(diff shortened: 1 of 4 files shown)


mark-moir:
Thanks @lisandrasilva, nice work.  We should be able to accept this soon, but there are a few comments we would like you address first, please.

LibraBFT/Abstract/BFT.agda
```diff
@@ -6,11 +6,479 @@
...
+
+
+   span-hon : ∀ {xs dis hon : List Member} {x : Member}
+              → span Meta-dishonest? xs ≡ (dis , x ∷ hon)
+              → x ∈ xs ×  Meta-Honest-PK (getPubKey x)
+   span-hon {x₁ ∷ []} {dis} {hon} {x} eq
+     with Meta-dishonest? x₁
+   span-hon {x₁ ∷ []} {dis} {hon} {x} () | yes prf
+   span-hon {x₁ ∷ []} {[]} {[]} {x₁} refl | no ¬dis = here refl , ¬dis
+   span-hon {x₁ ∷ xs} {dis} {hon} {x} eq
```
mark-moir, line 429: There is a catchall warning here, because there is no reason why `xs` can't be `[]`, which is covered by the previous cases as well.  Generally, we like to avoid warnings unless there is a good (documented) reason, or at least a `TODO-n` comment with some explanation of why it's there, so hopefully someone can address it later.

However, looking closer at this proof to see if I could easily remove the warning, I found that the proof is not as readable as most of your proofs.  I made another proof that doesn't have any warnings, and is more readable (to me, at least), though it's not perfect.  I'm pasting it below, feel free to take it or modify it if you like.

A few things I did here:
* Factored out a generic version of the lemma I needed, which makes it less cluttered with application-specific noise, and potentially enables reuse in other contexts.  I think there are a number of your proofs that would benefit from similar treatment.  Sufficiently generic lemmas should go in `Lemmas.agda`.
* Made the induction mirror that of the definition of `span`.   I think this is a big part of the reason it's simpler.

```
   -- Surely there is a library property?  If not, this should go in Lemmas
   []≢a∷as : ∀ {ℓ : Level} {A : Set ℓ}{a : A}{as1 as2 : List A}
           → as1 ≡ [] → as1 ≡ a ∷ as2
           → ⊥
   []≢a∷as {as1 = []} refl ()

   span-correct-1 : ∀ {ℓ₁ ℓ₂ A}{P : Unary-Pred {ℓ₁} A ℓ₂}{P? : Unary-Decidable P}{xs rest : List A}{x}
                  → proj₂ (span P? xs) ≡ x ∷ rest
                  → x ∈ xs × ¬ (P x)
   span-correct-1 {P? = P?} {[]} ()
   span-correct-1 {P? = P?} {xs = x ∷ xs} eq
     with does (P? x) | Dec.proof (P? x)
   ...| false         | ofⁿ ¬p rewrite ∷-injectiveˡ eq = here refl , ¬p
   ...| true          | _
      with span P? xs | inspect (span P?) xs
   ...| (_ , [])      | _ = ⊥-elim ([]≢a∷as refl eq)
   ...| (_ , _ ∷ _)   | [ eq1 ] rewrite ∷-injectiveˡ eq
      = ×-map₁ there (span-correct-1 (cong proj₂ eq1))

   span-hon : ∀ {xs dis hon : List Member} {x : Member}
            → span Meta-dishonest? xs ≡ (dis , x ∷ hon)
            → x ∈ xs × Meta-Honest-PK (getPubKey x)
   span-hon = span-correct-1 ∘ (cong proj₂)
```

LibraBFT/Abstract/BFT.agda
```diff
@@ -6,11 +6,479 @@
...
+   ... | fst , [] | [ eq₁ ] = cong suc (span-dis {xs} eq₁)
+
+
+
+   find-honest : ∀ {xs : List Member}
+                    → IsSorted _<Fin_ xs
+                    → bizF + 1 ≤ length xs
+                    → ∃[ α ] (α ∈ xs × Meta-Honest-PK (getPubKey α))
+   find-honest {xs} sxs biz<
+     with span Meta-dishonest? xs | inspect (span Meta-dishonest?) xs
```
mark-moir, line 458: I am curious as to why you used `span` here.  I suspect this could be simpler, but am out of time to think more about how.  Easy to say, right :smile:.

LibraBFT/Abstract/BFT.agda
```diff
@@ -6,11 +6,479 @@
...
+  (authorsN  : ℕ)
+  (bizF      : ℕ)
+  (isBFT     : authorsN ≥ suc (3 * bizF))
+  (getPubKey : Fin authorsN → PK)
+
+
+
+ where
 
   -- This is a utility function to make it easy to provide the bft-assumption
```
mark-moir, line 22: Please update the comment to make clear that the purpose of the module is `bft-lemma` at the end.

mark-moir (changes requested):
This is nice work!  I've made a few comments that I think can help to improve it.  I am out of time for today, but I think it's worthwhile you revising this and I will review again.  Thanks.

LibraBFT/Impl/Consensus/Types/EpochDep.agda
```diff
@@ -147,7 +147,8 @@ module LibraBFT.Impl.Consensus.Types.EpochDep (𝓔 : EpochConfig) where
   record IsValidQC (qc : QuorumCert) : Set where
     field
       ₋ivqcVotesValid      : All (IsValidVote ∘ rebuildVote qc) (qcVotes qc)
-      ₋ivqcIsQuorum        : IsQuorum {!!}  -- TODO: extract list of abstract members using ₋ivqcVotesValid?
+      ₋ivqcIsQuorum        : IsQuorum (All-reduce ₋ivvMember ₋ivqcVotesValid)
+                             -- TODO: extract list of abstract members using ₋ivqcVotesValid?
```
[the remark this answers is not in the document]
lisandrasilva: Done!

LibraBFT/Abstract/BFT.agda
```diff
@@ -6,11 +6,479 @@
...
+
+ -- The set of members of this epoch.
+ Member : Set
+ Member = Fin authorsN
+
+ Meta-dishonest? :  ∀ (m : Member) → Dec (Meta-Dishonest-PK (getPubKey m))
+ Meta-dishonest? m = Meta-DishonestPK? (getPubKey m)
+
+ module _  (bft-assumption : ∀ {xs : List Member} → IsSorted _<Fin_ xs
+                             → length (List-filter Meta-dishonest? xs) ≤ bizF)
```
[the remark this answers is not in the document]
lisandrasilva: Done! I also changed the formatting everywhere in the file.

LibraBFT/Abstract/BFT.agda
```diff
@@ -6,11 +6,479 @@
...
+   pred-Fin (suc x) = x
+
+
+   head-Sort : ∀ {n x l} {xs : List (Fin n)} → IsSorted _<Fin_ (x ∷ xs)
+         → length (x ∷ xs) ≡ l → l ≤ n → toℕ x ≤ n ∸ l
+   head-Sort {suc zero} {zero} {suc zero} {[]} _ _ _ = z≤n
+   head-Sort {suc zero} {zero} {suc l} {zero ∷ xs} (on-∷ () ∷ sxs) l≡ _
+   head-Sort {suc (suc n)} {zero} {suc 0} {[]} sxs refl _ = z≤n
+   head-Sort {suc (suc n)} {suc x} {suc 0} {[]} sxs refl _ = s≤s (head-Sort ([] ∷ []) refl (s≤s z≤n))
+   head-Sort {suc (suc n)} {x} {suc (suc l)} {x₃ ∷ xs} (on-∷ x₁ ∷ sxs) l≡ p
```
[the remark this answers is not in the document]
lisandrasilva: Done! And I also simplified so we have one case less than we had before.

LibraBFT/Abstract/BFT.agda
```diff
@@ -6,11 +6,479 @@
...
+
+ module _  (bft-assumption : ∀ {xs : List Member} → IsSorted _<Fin_ xs
+                             → length (List-filter Meta-dishonest? xs) ≤ bizF)
+   where
+
+   intersectElem : List Member → Member → List Member
+   intersectElem [] y = []
+   intersectElem (x ∷ xs) y
+      with Fin-<-cmp x y
+   ... | tri< a ¬b ¬c = intersectElem xs y
```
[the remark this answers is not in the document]
lisandrasilva: Done.

LibraBFT/Abstract/BFT.agda
lisandrasilva: It can't be empty because we have `span Meta-dishonest? xs ≡ (dis , x ∷ hon)`, therefore `xs` must have at least `x`. 
I simplified a little, I thought we needed to prove the base case in order to do induction but we don't. Check if it's ok. But if you still prefer your solution I can change.

Regarding the warnings I am not getting them, we can check this out later.

LibraBFT/Abstract/BFT.agda
lisandrasilva: Done.

LibraBFT/Abstract/BFT.agda
lisandrasilva: It seemed simpler to me. An alternative would be: first prove that the `length (List-filter Meta-dishonest? xs) ≤ length xs` and then prove that if it's `_<_` then `∃[ α ] (α ∈ xs × Meta-Honest-PK (getPubKey α))`. Otherwise, that is, if `length xs ≡ length (List-filter Meta-dishonest? xs)` we get a contradiction using the bft assumption (as we have now). I need to prove 2 auxiliary lemmas anyway. But at the moment using the span was more intuitive to me.

LibraBFT/Abstract/BFT.agda
mark-moir: About the warning, you probably want to customize `agda2-highlight-catchall-clause-face` in emacs to make these warnings visible.

LibraBFT/Abstract/BFT.agda
mark-moir: That is how I would have approached it (but proving more general versions of those lemmas; in fact, the first one is in the library as `length-filter`).  But your new version of the proof of `span-hon` is much nicer and simpler (than both your initial version and my version I gave in feedback), so let's stick with this for now.

mark-moir: But can I ask you one more thing: put a `TODO-1` comment suggesting that the proof can likely be simplified and briefly describe what you (and I) said above?  Once you do that, I think we can merge.  Thanks!

lisandrasilva: Done! Thank you for the feedback!

mark-moir:
Great, thanks.  Merging!

PR #19, opened 2021-03-24
Proved postulates on System model

lisandrasilva:
Proved the following postulates on the `LibraBFT.Yasm.System`:
- `cheatStepDNMPeerStates`
- `msgs-stable`
- `sendMessages-target`
- `peersRemainInitialized`

In order to prove the last one I needed to add a new parameter to the `SystemParameters` which is a `Decidable` over `PeerId`. Actually, I think that this new parameter will also be useful for some other lemmas that we may need to prove in future.

### DIFF
--- a/LibraBFT/Yasm/Base.agda
+++ b/LibraBFT/Yasm/Base.agda
@@ -45,6 +45,9 @@ module LibraBFT.Yasm.Base
     -- Finally, messages must carry an epoch id and might have an author
     part-epoch  : Part → EpochId
 
+    -- A decidable over PeerId's
+    _≟Peer_ : ∀ (p₁ p₂ : PeerId) → Dec (p₁ ≡ p₂)
+
     -- Initializes a potentially-empty state with an EpochConfig
     init : PeerId → EpochConfig → Maybe PeerState → PeerState × List Msg
 
--- a/LibraBFT/Yasm/Base.agda
+++ b/LibraBFT/Yasm/Base.agda
@@ -32,6 +32,7 @@ module LibraBFT.Yasm.Base
   constructor mkSysParms
   field
     PeerId    : Set
+    _≟PeerId_ : ∀ (p₁ p₂ : PeerId) → Dec (p₁ ≡ p₂)
     PeerState : Set
     Msg       : Set
     Part      : Set -- Types of interest that can be represented in Msgs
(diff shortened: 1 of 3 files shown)


LibraBFT/Yasm/Base.agda
```diff
@@ -45,6 +45,9 @@ module LibraBFT.Yasm.Base
     -- Finally, messages must carry an epoch id and might have an author
     part-epoch  : Part → EpochId
 
+    -- A decidable over PeerId's
+    _≟Peer_ : ∀ (p₁ p₂ : PeerId) → Dec (p₁ ≡ p₂)
```
mark-moir, line 49: Funnily enough, just yesterday I did the same in the branch I am working on.  Could you move this to be the second parameter (right after `PeerId`), which will avoid a conflict with my similar change.  Also, please spell it like this:
```
  _≟PeerId_ : (p1 p2 : PeerId) → Dec (p1 ≡ p2)
```

LibraBFT/Yasm/System.agda
```diff
@@ -250,23 +251,38 @@ module LibraBFT.Yasm.System
...
-   sendMessages-target : ∀ {m : SenderMsgPair} {sm : SentMessages} {ml : List SenderMsgPair}
-                       → ¬ (m ∈ sm)
-                       → m ∈ (ml ++ sm)
-                       → m ∈ ml
+
+ msgs-stable : ∀ {e e'} {pre : SystemState e} {post : SystemState e'} {m}
+             → (theStep : Step pre post)
+             → m ∈ msgPool pre
+             → m ∈ msgPool post
+ msgs-stable (step-epoch _)   m∈ = m∈
```
mark-moir, line 259: Please remove the extraneous space before the second parameter.

mark-moir (changes requested):
Thanks @lisandrasilva, looks good.  Before I can merge, I need you to send one more commit with signoff.  I have made some super minor suggestions to include in the commit.  Thanks!

LibraBFT/Yasm/Base.agda
lisandrasilva: Done!

mark-moir (approved):
Looks good.   Thanks @lisandrasilva.

PR #21, opened 2021-03-31
Make `peerStates` total, eliminate pain dealing with `Maybe PeerState`

mark-moir:
This is an experimental change to our system model, in which `peerStates` becomes a total function from `PeerId` to `PeerState`, meaning that even peers that don't exist yet have a "state".

This change eliminates having to deal with `Maybe PeerState` all the time, but also we cannot prove anything about the state of an uninitialised peer and we cannot distinguish them from initialised ones.

To address this issue, we also introduce a new element to `SystemState` called `initialised`, which tracks which peers' states have been initialised.  We can then state/prove properties about initialised peer states.  An example is the (postulated, for now) `qcVotesSentB4` property, which is required to hold only for initialised peers.

There are also a few minor unrelated changes.  For one example, the bridge between the system model and our (fake) handler paired the identity of the peer taking the step and the message it is processing, which is misleading as it suggests that the first element of the pair is the sender of the message, which is incorrect.

### DIFF
--- a/LibraBFT/Concrete/Properties/VotesOnce.agda
+++ b/LibraBFT/Concrete/Properties/VotesOnce.agda
@@ -43,10 +43,10 @@ module LibraBFT.Concrete.Properties.VotesOnce where
 
  ImplObligation₁ : Set₁
  ImplObligation₁ =
-   ∀{e pid pid' s' outs pk}{pre : SystemState e}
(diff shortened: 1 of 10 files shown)


mark-moir:
Out-of-band feedback is positive, so I am merging.

PR #23, opened 2021-04-08
Proved ImplObligation1 for the fake handler without using the unwind 

lisandrasilva:
Proved `ImplObligation₁` for the VotesOnlyOnce property for the fake handler. There is already a version of this proof, however uses the unwind approach. We managed to prove it without using the unwind making the proof significantly smaller. 

We also did a small simplification on the `ImplObligation₂`. 

As the fake handler will get more and more complex, we decided that both proofs can be useful in the future. Therefore, we keep both approaches to prove the `ImplObligation₁` in the repo, although in different files. The new file `VotesOnceDirect` contains the version of the proof without the unwind.


LibraBFT/Impl/Properties/VotesOnceWU.agda
```diff
@@ -0,0 +1,102 @@
...
+open        Structural impl-sps-avp
+open import LibraBFT.Impl.Properties.VotesOnce
+
+-- In this module, we (will) prove the two implementation obligations for the VotesOnce rule.  Note
+-- that it is not yet 100% clear that the obligations are the best definitions to use.  See comments
+-- in Concrete.VotesOnce.  We will want to prove these obligations for the fake/simple
+-- implementation (or some variant on it) and streamline the proof before we proceed to tacke more
+-- ambitious properties.
+
+module LibraBFT.Impl.Properties.VotesOnceWU where
```
mark-moir, line 34: The name is a bit cryptic.  I guess WU is "Without Unwind"?  Maybe replace "WU" with "Direct"?  Please also put a comment in the top of this file mentioning that it duplicates the original proof, and we are keeping both for now to highlight the different approaches.

LibraBFT/Yasm/Properties.agda
```diff
@@ -78,6 +78,12 @@ module LibraBFT.Yasm.Properties
    trans (sym (vp-ec-≡ vp1))
          (trans (AE.lookup-𝓔s-injective 𝓔s (vp-epoch vp1) (vp-epoch vp2) parts≡)
                 (vp-ec-≡ vp2))
+ postulate
```
mark-moir, line 81: Please add a `TODO-1 : prove it` comment (see the [style guide](https://github.com/oracle/bft-consensus-agda/blob/main/STYLE.md#assumptions-warnings-and-overrides)).  However, I think it's reasonable to keep this as a postulate for now as it is likely to become obsolete soon, when the moral equivalent of `ValidSenderForPK` will become an application-provided module parameter, so assuming we still need this property, it will need to be provided by the application too.

LibraBFT/Yasm/System.agda
```diff
@@ -252,6 +252,12 @@ module LibraBFT.Yasm.System
                         → peerStates (StepPeer-post theStep) ≡ peerStates pre
  cheatStepDNMPeerStates {pid = pid} {pre = pre} (step-cheat _ _) _ = override≡Correct {f = peerStates pre} {pid}
 
+ cheatStepDNMPeerStates₁ : ∀{e pid pid' st' outs}{pre : SystemState e}
+                        → (theStep : StepPeer pre pid st' outs)
+                        → isCheat theStep
+                        → peerStates (StepPeer-post theStep) pid' ≡ peerStates pre pid'
+ cheatStepDNMPeerStates₁ {_} {pid} {pid'} (step-cheat fm x₁) x = overrideAux pid pid'
```
mark-moir, line 259: If we're going to use `overrideAux` directly, let's give it a better name.  Maybe `overrideSameVal-correct`, and rename `override≡Correct` to `overrideSameVal-correct-ext` (the extensional version of the same property)?

LibraBFT/Impl/Properties/VotesOnceWU.agda
```diff
@@ -0,0 +1,102 @@
...
+                  in ≤-trans (oldVoteRound≤lvr r pidIn pkH sig msb4 vspk ep≡v) lvr≤
+  oldVoteRound≤lvr {pid = pid'} {pre = pre}
+                   step@(step-s r (step-peer {pid = pid} stHon@(step-honest stPeer)))
+                   pidIn pkH sig msv vspk ep≡
+     | inj₁ (m∈outs , vspkN , newV)
+     with sameHonestSig⇒sameVoteData pkH (msgSigned msv) sig (msgSameSig msv)
+  ...| inj₁ hb = ⊥-elim (PerState.meta-sha256-cr pre step hb)
+  ...| inj₂ refl
+     with sameEpoch⇒sameEC vspk vspkN refl
+          -- Both peers are allowed to sign for the same PK, so they are the same peer
```
mark-moir, line 70: Add "in the same epoch" after "same PK".

mark-moir (changes requested):
Hi @lisandrasilva, this is great, thanks!  I only have relatively minor comments.

PR #24, opened 2021-04-15
Move towards more realistic initialisation, banish `EpochConfig` from `Yasm`

mark-moir:
This pull request moves towards more realistic initialisation and management of `EpochConfig`s.  There are several aspects:

### Banish `EpochConfig`s from `Yasm`

We remove everything related to `EpochConfig` (which is an application specific notion) from the system model (`Yasm`), instead defining a type that the application can use to express which peers can sign messages for which (honest) public keys. 
 `Yasm` now accepts module parameters defining a `GenesisInfo` type and a value of that type that is provided to `step-init`.  Then we propagate these changes through the repo.

Also cleaned up some obsolete and confusing aspects of `Yasm`:
- eliminated the function cheat steps use to determine a message to send (they can send any message they want, as long as it satisfies the constraint on forging signatures, there is no need for a function to justify the choice of message)
- eliminated obsolete parameter to `StepPeerState` which was originally intended to update `initialisation`, but this is handled in `StepPeer-post`

### ~~Make individual peer states track (meta) `EpochConfig`s they know about~~

~~These are bundled together with an actual peer state in type `EventProcessorAndMeta`.  This change required making `Yasm` support peer states at arbitrary levels, and similarly required making `RWST`/`LBFT` for arbitrary levels; this required commenting out lens-related functionality (not yet used) in `RWST` because our lenses do not yet support arbitrary levels.  As part of this, we define various constants to better document the reasons for various levels.~~

### Make `ValidSenderForPK` a function of `SystemState` (rather than `PeerState`)

This change enables the evidence that a peer is allowed to sign for a given PK in a given epoch to be derived from `SystemState`, rather than duplicating it in every peer's state.

### Make fake implementation initialisation slightly more realistic

We make some steps towards more realistic initialization from `GenesisInfo`, but postulating parts of it and some temporary, bogus properties.

### Fix proofs

We propagate the changes to `Yasm` throughout all the proofs.  Proving `ImplObligation₁` required changing its definition because, with the changes to `Yasm`, it is now possible that a single peer step (`step-init` or `step-msg`) can learn of a new `EpochConfig` and therefore know that it can sign a message for its public key in that epoch *and* sign such a message.  The fact that it can do so ~is recorded in the peer's state~ may become true as a result of this step (e.g., if a peer learns that an epoch change has occurred and then signs a message for that epoch in the same step), and does not hold in its prestate.

### Explore broken "unwind-free" proof

This proof was broken by the changes to `Yasm` and requires more work to bring it up to date, in part because of the above-described change to the proof obligation.  Some parts are commented out and some holes made to explore the issues, and a comment is added at the top of `VotesOnceDirect.agda`.


haroldcarr:
Question regarding : "Banish EpochConfigs from Yasm"

The description says EpochConfig things have been removed and replaced with `GenesisInfo` and `step-init`.
I do not see those new things.
What I do see is that epoch change parameters are passed in.
I do not see `GenesisInfo` in this branch.
Please clarify.

haroldcarr:
Question regarding  "... which peers can sign messages for which (honest) public keys."

Do messages contain both epoch and round?

When "receiving" a message, is the epoch of the message checked against the epoch of the receiving peer?

mark-moir:
@haroldcarr , for your first question regarding `GenesisInfo`, maybe you're on the wrong branch?  Search for `GenesisInfo` and `genInfo` and you will find the relevant things.

Regarding

> Do messages contain both epoch and round?

I might be misunderstanding your question, as I know that you know already know the message types in intimate detail.  But yes, all messages have an associated epoch (and round, not that the latter is relevant to "which peers can sign messages for which (honest) public keys").  For example, a `ProposalMessage` contains a `Block`, which contains `BlockData`, which has epoch and round. 

Regarding

> When "receiving" a message, is the epoch of the message checked against the epoch of the receiving peer?

These are details of the implementation.  The answer for the actual implementation (that we will model in future) is "yes" (as you know better than I do), but our current fake/trivial implementation does not do those things.

HTH.

haroldcarr:
Thanks for the clarifications.
And I guess I was on the wrong branch, but I swear I checked more than once.

mark-moir:
Closing as this pull request is superseded by #30.

PR #26, opened 2021-04-27
rename types

haroldcarr:
```
rename :
- LockedRound       -> PreferredRound
- EpochId           -> Epoch
- PersistentStorage -> PersistentSafetyStorage
- ProcessedVMOutput -> StateComputeResult

Note: when renaming, variables and acronyms were renamed consistently also.

refactored stuff from
- Persistent[Saftey]Storage to SafetyData

change constructor names
- mk* -> *.new (e.g., mkBlock -> Block.new where the '.' is a solid bullet)
```

### DIFF
--- a/LibraBFT/Impl/Consensus/Types/EpochIndep.agda
+++ b/LibraBFT/Impl/Consensus/Types/EpochIndep.agda
@@ -45,7 +45,7 @@ module LibraBFT.Impl.Consensus.Types.EpochIndep where
   -----------------
 
   record BlockInfo : Set where
-    constructor mkBlockInfo
+    constructor BlockInfo∙new
     field
       ₋biEpoch : EpochId
       ₋biRound : Round
@@ -57,11 +57,11 @@ module LibraBFT.Impl.Consensus.Types.EpochIndep where
   postulate instance enc-BlockInfo : Encoder BlockInfo
 
   BlockInfo-η : ∀{e1 e2 r1 r2 i1 i2} → e1 ≡ e2 → r1 ≡ r2 → i1 ≡ i2
-              → mkBlockInfo e1 r1 i1 ≡ mkBlockInfo e2 r2 i2
+              → BlockInfo∙new e1 r1 i1 ≡ BlockInfo∙new e2 r2 i2
   BlockInfo-η refl refl refl = refl
 
   record LedgerInfo : Set where
-    constructor mkLedgerInfo
+    constructor LedgerInfo∙new
     field
       ₋liCommitInfo        : BlockInfo
       ₋liConsensusDataHash : HashValue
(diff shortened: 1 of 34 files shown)


LibraBFT/Impl/Consensus/Types/EpochIndep.agda
```diff
@@ -257,12 +257,22 @@ module LibraBFT.Impl.Consensus.Types.EpochIndep where
              (siHighestQuorumCert ∷ siHighestCommitCert ∷ [])
   postulate instance enc-SyncInfo : Encoder SyncInfo
 
+  {- TODO-HC how do if-then-else and equality check?
```
mark-moir, line 260: Please change `HC-TODO` to `TODO-1` (see the [`TODO` guide](https://github.com/oracle/bft-consensus-agda/blob/main/TODO.md)).  Also, here are some suggestions to help you make progress on the `TODO`, if you like.

Unlike Haskell, we cannot simply say `deriving Eq`; we need to define equality for `QuorumCert`.  That is straightforward and slightly tedious.  I've postulated it below (`_≟QC_`), and then shown how we can use it to define what you want.
```
  record SyncInfo : Set where
    field
      ₋siHighestQuorumCert  : QuorumCert
      ₋siHighestCommitCert  : Maybe QuorumCert
      -- ₋siHighestTimeoutCert : Mabe TimeoutCert -- Not used yet.
  open SyncInfo public
  unquoteDecl siHighestQuorumCert   = mkLens (quote SyncInfo)
             (siHighestQuorumCert ∷ [])
  postulate instance enc-SyncInfo : Encoder SyncInfo

  postulate -- TODO-1: define this, it should be easy, but requires defining Dec instances for Vote
            -- Data, BlockInfo, and KVMap.
    _≟QC_ : (q1 q2 : QuorumCert) → Dec (q1 ≡ q2)

  _QCBoolEq_ : QuorumCert → QuorumCert → Bool
  _QCBoolEq_ q1 q2 = does (q1 ≟QC q2)

  SyncInfo∙new : QuorumCert → QuorumCert → SyncInfo
  SyncInfo∙new highestQuorumCert highestCommitCert =
    record { ₋siHighestQuorumCert = highestQuorumCert
           ; ₋siHighestCommitCert = if highestQuorumCert QCBoolEq highestCommitCert
                                    then nothing else (just highestCommitCert) }

  siHighestCommitCert : Lens SyncInfo QuorumCert
  siHighestCommitCert = mkLens' (λ x → maybe id (x ^∙ siHighestQuorumCert) (₋siHighestCommitCert x))
                                (λ x si → record x { ₋siHighestCommitCert = just si })

```
Note that I removed the "automatic" lens for `siHighestCommitCert` so I could replace it with the manual one above.  Also I made `₋siHighestCommitCert  : Maybe QuorumCert`, which I think is what we want, right?

mark-moir (changes requested):
Thanks @haroldcarr.  Looks good.  I have added a comment with some suggestions for the `TODO` you left.  Also please note:

- remember to "signoff" your last commit
- please update the pull request description with a little more detail than the title, so we can remember what this about in future (e.g., mention versions of LBFT and maybe summarize the main changes).

Thanks!

mark-moir:
Thanks @haroldcarr, merging now.

LibraBFT/Impl/Consensus/Types/EpochIndep.agda
mark-moir: Actually, I found a more convenient way using the (postulated, for now) fact that we have an injective encoder for `QuorumCert`; see #27.

PR #27, opened 2021-04-29
Add decidable equality for any type with an `Encoder`, use it to implement conditional `SyncInfo` constructor and lens

mark-moir:
The constructor for `SyncInfo` compares two `QuorumCert`s and constructs the `SyncInfo` dependent on the outcome.  We observe that `QuorumCert` has (postulated, for now) an injective encoding, and we can implement decidable equality for any such type.

This pull request provides support for that and then uses it to implement the constructor and lens for `SyncInfo`.  This avoids the need to manually implement decidable equality  for `QuorumCert`, `VoteData`, `BlockInfo`, `LedgerInfoWithSignatures`, etc.

### DIFF
--- a/LibraBFT/Base/ByteString.agda
+++ b/LibraBFT/Base/ByteString.agda
@@ -60,6 +60,9 @@ module LibraBFT.Base.ByteString where
   ByteString : Set
   ByteString = List (Vec Bool 8)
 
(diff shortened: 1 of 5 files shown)


haroldcarr:
Reviewed live.

PR #30, opened 2021-05-06
Banish `EpochConfig` from Yasm and move towards more realistic initialisation

mark-moir:
This pull request supersedes #24.  It is the culmination of work done on branch `mark-begin-initialisation` as seen in #24, but is organised as a small number of (fairly) coherent commits.

This pull request moves towards more realistic initialisation and management of `EpochConfig`s.  There are several aspects:

### Banish `EpochConfig`s from `Yasm`

We remove everything related to `EpochConfig` (which is an application-specific notion) from the system model (`Yasm`), instead defining a type `ValidSenderForPK` that the application can use to express which peers can sign messages for which (honest) public keys in a given `SystemState`.

 `Yasm` now accepts module parameters defining a `GenesisInfo` type and a value of that type that is provided to `step-init`.  Then we propagate these changes through the repo.

Also cleaned up some obsolete and confusing aspects of `Yasm`:
- eliminated the function cheat steps use to determine a message to send (they can send any message they want, as long as it satisfies the constraint on forging signatures, there is no need for a function to justify the choice of message)
- eliminated obsolete parameter to `StepPeerState` which was originally intended to update `initialisation`, but this is handled in `StepPeer-post`

### Make fake implementation initialisation slightly more realistic

We make some steps towards more realistic initialization from `GenesisInfo`, but postulating parts of it.

### Fix proofs

We propagate the changes to `Yasm` throughout all the proofs.  Proving `ImplObligation₁` required changing its definition because, with the changes to `Yasm`, it is now possible that a single peer step (`step-init` or `step-msg`) can learn of a new `EpochConfig` and therefore know that it can sign a message for its public key in that epoch *and* sign such a message.  The fact that it can do so may become true as a result of this step (e.g., if a peer learns that an epoch change has occurred and then signs a message for that epoch in the same step), and does not hold in its prestate.

### Explore broken "unwind-free" proof

This proof was broken by the changes to `Yasm` and requires more work to bring it up to date, in part because of the above-described change to the proof obligation.  Some parts are commented out and some holes made to explore the issues, and a comment is added at the top of `VotesOnceDirect.agda`.


mark-moir:
This pull request supersedes #24.  It is the culmination of work done on branch `mark-begin-initialisation` as seen in #24, but organised as a small number of (fairly) coherent commits.

PR #32, opened 2021-05-12
Enable an initial "genesis" message at system initialization, make unwind work for non-genesis messages, fix properties

mark-moir:
In this pull request, we model the initial distribution of some information to all participants by making the system model accept an initial message that is placed in the message pool in the initial state.  This ensures that the model allows cheat steps to send messages with signatures that can be derived from this initial information.

One consequence is that we cannot `unwind` a message containing parts from `GenesisInfo` because they are not introduced by any step.  This is handled by introducing properties about initial vs. non-initial parts (in our case, `Vote`s represented in the `GenesisInfo` are all for round 0, while new `Vote`s sent by peer steps are always for non-zero rounds.

We also postulate/prove various `GenesisInfo` for the (still fake) implementation model and properties about it required to make all the proofs go through, and then do so.

I have also experimented with an alternative in which the system model accepts a separate `GenesisInfo` type, which requires changing `CheatConstraint` to reflect that cheats can behave in this way.  The two approaches turn out to be fairly similar, although I have not completed the latter (yet?).  I am sharing this one for comments and feedback.

One downside of this approach is that it is no longer the case that the initial message pool is empty, which means some `step-0` cases require proofs that didn't before.  The alternative approach would instead result in additional cases for cheat steps, saying that the part was in the `GenesisInfo`, which will then have to be eliminated as contradictions.  In hindsight, that may be preferable because the (new) `step-0` proofs require reasoning about `msgPool initialState`.

### DIFF
--- a/LibraBFT/Base/PKCS.agda
+++ b/LibraBFT/Base/PKCS.agda
@@ -158,6 +158,16 @@ module LibraBFT.Base.PKCS where
  ...| no ¬Signed = ⊥-elim (¬Signed (isSigned wvs1))
  ...| yes sc = withVerSig-≡ {wvs1 = wvs1} {wvs2 = wvs2} (Signed-pi c (isSigned wvs1) (isSigned wvs2))
 
+ -- A type for expressing that a property that holds for one C holds for any other C with the same
+ -- signature.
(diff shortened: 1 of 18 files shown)


mark-moir:
I'm closing this as work on the alternative approach mentioned in the description, along with some simplifications that could probably apply to this approach as well, looks more promising.

PR #33, opened 2021-05-13
Finished proof of `VotesOnceDirect`

mark-moir:
This pull request supersedes #31; this is just addressing merge conflicts.

Finished proof of VotesOnceDirect with the new data type PeerCanSignForPK. Also proved the auxiliary lemmas needed for the proof. Note that the auxiliary lemma MsgWithSig⇒ValidSenderInitialised is not used yet but will probably be useful in the future to prove peerCanSign-Msb4 when we address epoch change transitions.


mark-moir:
This is work by @lisandrasilva, see #31, which this pull request supersedes.  I just helped with the merge.

PR #34, opened 2021-05-26
Introduce "genesis information", modify system model to reflect that cheaters can send "new" signatures from it 

mark-moir:
The main purpose of this pull request is to extend the system model with a notion of "genesis information" that is passed to initialising peers, and relax the constraint on cheat steps to reflect that a `Signature` represented in this genesis information can be sent by a cheater, even if it has not been "sent" before, and to fix the proofs (weakening them to be only about signatures that are not in genesis information in some cases) to accommodate this change.

Some other pieces included are:
* Updating some implementation types to match the more recent implementation
* Simplify/improve the (fake) `processProposalMsg`
* Eliminating an inconsistent and unnecessary attempt at generality regarding collision resistant hash functions
* Breaking `VotesOnceDirect` yet again, rumour is that someone ~might fix it soon~ [has fixed it](#35 ) :smile:.
* Move a few properties to better places, including eliminating `LibraBFT.Impl.Properties.Aux`

Credits: Some changes from @haroldcarr (more modern types) and @lisandrasilva (much simpler version of a proof than my first attempt) and @cwjnkins (proving a postulate and simplifying `processProposalMsg`) have been included out-of-band.


mark-moir:
After "self review" and addressing my own comments, I am merging.

PR #35, opened 2021-05-26
Proved `VotesOnceDirect`

lisandrasilva:
After the changes in initialisation the proof of `VotesOnceDirect` on implementation side was broken. In this pull request the proof of `VotesOnceDirect` was fixed. therefore we have again and inductive proof of `VotesOnlyOnce` rule.

### DIFF
--- a/LibraBFT/Base/KVMap.agda
+++ b/LibraBFT/Base/KVMap.agda
@@ -24,7 +24,8 @@ open import LibraBFT.Prelude
 
 module LibraBFT.Base.KVMap  where
 
- variable
+ private
+  variable
    Key : Set
    Val : Set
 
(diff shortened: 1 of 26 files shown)


mark-moir:
Thanks @lisandrasilva.  Looks good, just a few minor comments.
* The correct module comment is in the wrong place, and there is an outdated comment where it should be.  So please remove the comment at lines 32-36 and move the one from lines 8-10 there.
* A number of property definitions don't follow formatting conventions, perhaps due to changing names (e.g., `newVoteEpoch≡⇒Round≡`, `peerCanSignEp≡`, `peerCanSign-Msb4`. ..).
* Line 129 could be `...| refl = refl`.
I'm happy to merge this if you address those comments.  Thanks!

mark-moir:
Thanks @lisandrasilva!  Merging now.

PR #40, opened 2021-06-03
Define implementation obligations for `PreferredRound` rule

mark-moir:
- The main purpose of this pull request is to define the implementation obligations related to the `PreferredRound` rule (`LibraBFT.Concrete.Properties.PreferredRound`), so that, assuming an implementation complies with these obligations, we can prove `LibraBFT.Concrete.Obligations.PreferredRound.Type` for our intermediate system state, and we have already proved (`LibraBFT.Concrete.Obligations.PreferredRound.proof`) that this implies the corresponding assumption made by the abstract proofs.
- The new implementation obligations refer to structures such as `Cand-3-chain-vote` and `VoteParentData`, which are in terms of abstract `Record`s and therefore require an `EpochConfig`, which is introduced as a module parameter.  For consistency, the `VoteOnce` obligations now similarly receive an `EpochConfig` module parameter, although this is not strictly required.
- When defining the `PreferredRound` obligations using the the `VotesOnce` ones as a starting point, I noticed that the `PeerCanSignForPK` arguments were not actually needed (because the proof can use `impl-sps-avp` to get them based on other arguments), so I eliminated them.  Before realising this, however, I had some pain with these arguments, which was eased by refactoring `PeerCanSignFor`.  Even though I eliminated these arguments, I kept the refactoring as I think it may be helpful anyway.  I also left a trivial function `EC-member-cast` that I proved while struggling with the `PeerCanSignForPK` arguments before eliminating them, and similarly have left this even though it is not used now.
- As a sanity check, I started the proof in `LibraBFT.Impl.Properties.PreferredRound`, proving `PR.ImplObligation₂` using a proof almost identical to the corresponding one for `VotesOnce` (and left a `TODO` to refactor).  Note that our fake implementation does not comply with `PR.ImplObligation₁`, so nobody should try to prove that it does.  We will be working soon on a more realistic model that does comply.
- And a few minor things:
  - Fixed `availEpochsConsistent` to require the epoch IDs of the votes in question to be equal, otherwise the property will not hold in future when we introduce epoch changes.
  - Introduced `⊎-elimʳ` of type `¬ A₁ → A₀ ⊎ A₁ → A₀` (and its symmetric counterpart) and used it to streamline some proofs
  - Moved some properties to more appropriate places
  - Made `Yasm.Yasm` export `Yasm.Types` so it does not need to be explicitly imported everywhere.

### DIFF
--- a/LibraBFT/Abstract/Types/EpochConfig.agda
+++ b/LibraBFT/Abstract/Types/EpochConfig.agda
@@ -65,6 +65,12 @@ module LibraBFT.Abstract.Types.EpochConfig
                   → toNodeId 𝓔₁ mbr₁ ≡ toNodeId 𝓔₂ mbr₂
   PK-inj-same-ECs {𝓔₁} refl pks≡ = cong (toNodeId 𝓔₁) (PK-inj 𝓔₁ pks≡)
 
+  EC-member-cast : ∀ {𝓔₁ 𝓔₂ : EpochConfig}
+                 → 𝓔₁ ≡ 𝓔₂
+                 → Member 𝓔₁
+                 → Member 𝓔₂
+  EC-member-cast refl m = m
+
   module _ (ec : EpochConfig) where
     NodeId-PK-OK : PK → NodeId → Set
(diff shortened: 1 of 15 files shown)


cwjnkins (approved):
:+1:

PR #42, opened 2021-06-11
parts of process vote; PendingVotesSpec (not used)

haroldcarr:

This contains an incomplete implementation of `RoundManager.processVoteMsgM`.
If it is needed before I am available again, go ahead and complete it.  Otherwise, it is fine to leave it as is.

There are places in this request that have `IMPL-TODO` - it would be good to have feedback/fixes for those places.

### DIFF
--- a/LibraBFT/Impl/Consensus/RoundManager.agda
+++ b/LibraBFT/Impl/Consensus/RoundManager.agda
@@ -3,17 +3,24 @@
    Copyright (c) 2020, 2021, Oracle and/or its affiliates.
    Licensed under the Universal Permissive License v 1.0 as shown at https://opensource.oracle.com/licenses/upl
 -}
-open import Optics.All
-open import LibraBFT.Prelude
+
 open import LibraBFT.Base.ByteString
 open import LibraBFT.Base.PKCS
 open import LibraBFT.Base.Types
 open import LibraBFT.Hash
 open import LibraBFT.Impl.Base.Types
+open import LibraBFT.Impl.Consensus.BlockStorage.BlockStore   as BlockStore
+open import LibraBFT.Impl.Consensus.BlockStorage.SyncManager  as SyncManager
+open import LibraBFT.Impl.Consensus.ConsensusTypes.Vote       as Vote
+open import LibraBFT.Impl.Consensus.Liveness.ProposerElection as ProposerElection
+open import LibraBFT.Impl.Consensus.Liveness.RoundState       as RoundState hiding (processCertificatesM)
 open import LibraBFT.Impl.Consensus.Types
+open import LibraBFT.Impl.OBM.Prelude
 open import LibraBFT.Impl.Util.Crypto
 open import LibraBFT.Impl.Util.Util
 open import LibraBFT.Abstract.Types.EpochConfig UID NodeId
+open import LibraBFT.Prelude
+open import Optics.All
 
 
 -- This is a minimal/fake example handler that obeys the VotesOnce rule, enabling us to start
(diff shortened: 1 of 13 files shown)


LibraBFT/Impl/Consensus/RoundManager.agda
```diff
@@ -24,48 +31,126 @@ open import LibraBFT.Abstract.Types.EpochConfig UID NodeId
...
-  fakeBlockInfo eid rnd pm = BlockInfo∙new eid rnd (pm ^∙ pmProposal ∙ bId)
-
-  fakeLedgerInfo : BlockInfo → ProposalMsg → LedgerInfo
-  fakeLedgerInfo bi pm = LedgerInfo∙new bi (pm ^∙ pmProposal ∙ bId)
-
-  postulate -- TODO-1: these are temporary scaffolding for the fake implementation
-    fakeSK  : SK
-    fakeSig : Signature
-
-  processProposalMsg : Instant → ProposalMsg → LBFT Unit
```
mark-moir, line 45: I would like to keep our "fake" implementation (and proofs about it) for some time, rather than replacing them.  I guess we should move them somewhere else so they no longer pollute the namespace of the actual implementation model.  I think if you did that, the result would be the files the @cwjnkins is building on looking the same, thus avoiding conflicts for him, while establishing a new location for the fake handler.  Does this make sense?

LibraBFT/Impl/Consensus/Types.agda
```diff
@@ -1,9 +1,11 @@
+{-# OPTIONS --allow-unsolved-metas #-}
```
mark-moir, line 1: According to our [style guide](https://github.com/oracle/bft-consensus-agda/blob/main/STYLE.md#coding-style-conventions), the copyright notice should come before the pragma.

LibraBFT/Impl/Consensus/Types.agda
```diff
@@ -166,4 +233,17 @@ module LibraBFT.Impl.Consensus.Types where
...
+    g : RoundManager → SafetyRules
+    g rm = ₋rmEC rm ^∙ rmSafetyRules
+
+    s : RoundManager → SafetyRules → RoundManager
+    s rm sr = record rm { ₋rmEC = (₋rmEC rm) [ rmSafetyRules := sr ]}
+
+  lPersistentSafetyStorage : Lens RoundManager PersistentSafetyStorage
+  lPersistentSafetyStorage = lSafetyRules ∙ srPersistentStorage
 
+  lSafetyData : Lens RoundManager SafetyData
```
mark-moir, line 248: This name (and others like it above, I think) causes a disconnect with the Haskell code, which uses `pssSafetyData` here, which is derived using `obmMakeClassy`.  Not sure if it makes sense to try to emulate `obmMakeClassy`.  More pragmatically, maybe we could just establish a naming convention, so for example, we might use `pssSafetyData-RoundManager` or `pssSafetyData-derived` or something so we alert the reader to why it's different in each case.  Thoughts?

LibraBFT/Impl/Consensus/Types/EpochIndep.agda
```diff
@@ -528,3 +578,6 @@ module LibraBFT.Impl.Consensus.Types.EpochIndep where
     TooManySignatures    : Usize → Usize → VerifyError
     InvalidSignature     :                 VerifyError
 
+  -- TODO-1: Implement this (low priority)
+  ErrLog : Set
```
mark-moir, line 582: As discussed with @cwjnkins, something like this would be preferable:
```
data ErrLog : Set where
FakeError : ErrLog
```
This is one of multiple reasons that `unit` shows up in proofs, making them less readable (another is for our `RWST` environment defined for `LBFT` (in `LibraBFT.Impl.Util.Util`).  But maybe let's leave that until after we merge this and some of @cwjnkins work on top of it.

LibraBFT/Impl/Handle.agda
```diff
@@ -118,7 +122,7 @@ module LibraBFT.Impl.Handle where
  outputToActions : RoundManager → Output → List (LYT.Action NetworkMsg)
  outputToActions rm (BroadcastProposal p) = List-map (const (LYT.send (P p)))
                                                      (List-map proj₁
-                                                               (kvm-toList (-vvAddressToValidatorInfo (₋esVerifier (₋rmEpochState (₋rmEC rm))))))
+                                                               (kvm-toList (₋vvAddressToValidatorInfo (₋esVerifier (₋rmEpochState (₋rmEC rm))))))
```
mark-moir, line 125: Here again, I would like to maintain our fake handler while making progress on the real one.

mark-moir (changes requested):
Thanks @harold.  In general, I'm happy to merge this work in this state; it's WIP towards something that we will review in more detail later.  I do have a few comments.  It may make sense to address some of them after merging this and some of @cwjenkins ongoing work that builds on top of it.  However, ...

The most important issue that I think we should resolve before merging is that, rather than "evolving" the handler and implementation model from our fake one to a more real one, I think we should keep both, for example so we don't  break/lose proofs we have about the fake handler as we make progress on the real one.  If you could establish new locations for the fake stuff (so it doesn't continue to pollute the namespace rightfully occupied by the WIP real model), then we can keep the files you've changed in the same state, thus avoiding conflicts for @cwjnkins work that builds on it.  E.g., maybe establish `LibraBFT.FakeImpl`, which would contain at least:

* `Consensus/RoundManager.agda`
* `Consensus/RoundManager/Properties.agda`
* `Handle.agda`
* `Handle/Properties.agda`
* `Impl/Properties/VotesOnce.agda`
* `Impl/Properties/VotesOnceDirect.agda`
* `Impl/Properties/PreferredRound.agda`

And then fix the imports so these all work.

I also note that this will make even more conflicts than I already have for #43, so it may make sense to defer resolving those conflicts, and ask @lisandrasilva to make a new pull request after merging this and maybe @cwjnkins current work.

Does this make sense?

Any thoughts @cwjnkins ? @lisandrasilva ?

harold:
@mark-moir - I think you meant to @haroldcarr who, if I do say so myself, has a lovely first name.

Looks like a fun project, best of luck with the verification.

mark-moir:
@harold, yes indeed I meant @haroldcarr, thanks for the heads up.

cwjnkins:
@mark-moir I agree with your proposal to split off the fake and real implementation. I am at a point where it make sense to start connecting the work on proving properties about a single peer step to invariants of the system, and this split would make that much easier.

LibraBFT/Impl/Consensus/Types.agda
haroldcarr: FIXED

LibraBFT/Impl/Consensus/Types.agda
haroldcarr: all the `l*` lens are direct analogues of 
```
instance RWSafetyRules (RoundManager a) where
  lSafetyRules = lens _rmSafetyRules (\x y -> x { _rmSafetyRules = y})
```
There is NO difference (except in the way the Agda defines `LBFT` compared to the haskell version.

LibraBFT/Impl/Consensus/Types/EpochIndep.agda
haroldcarr: I was not privy to this conversation, so NO change for now.

LibraBFT/Impl/Consensus/Types.agda
mark-moir: "pssSafety" != "lSafetyRules".  That's a difference.  I am saying this causes a disconnect for anyone comparing the Haskell code and the Agda model thereof.  As mentioned in my comment, I'm not suggesting we try to make them identical, but rather establish a convention to make it easier on the reader to navigate.

mark-moir: But I see you have merged already, so let's keep this in mind for later.

LibraBFT/Impl/Consensus/Types/EpochIndep.agda
mark-moir: OK, I guess we'll keep this one for later too.

LibraBFT/Impl/Consensus/RoundManager.agda
mark-moir: I understand you'll do this in a subsequent PR.  I thought that your changes would break the existing proofs about the fake handler, which I wouldn't want to merge, but now I realise that you didn't actually hook up the partial implementation to the handler, so no proofs were harmed in the production of this PR :smile:.

LibraBFT/Impl/Consensus/Types.agda
haroldcarr: I do not think we are on the same page. 

`pssSafetyData` is a field inside `PersistentSafetyStorage`

`lSafetyRules` is a lens from `RoundManager` to `SafetyRules`

they are NOT the same type

mark-moir: How _could_ we be on the same page, given that you're on the wrong one?  😝   Unless I am of course, but I double checked and I believe:

* `_pssSafetyData` (note underscore) is a field inside `PersistentSafetyStorage`
* `pssSafetyData` is a lens generated by `obmMakeClassy` (I think the instance you quoted above is part of how it is generated, right?)

Maybe I should be more explicit about the disconnect I'm commenting on.  In `constructAndSignVoteM.continue2`, we have in the Haskell `pssSafetyData .= safetyData` whereas in the Agda model .... ah, actually, I mean the _future_ Agda model that @cwjnkins is already fleshing out in another branch :smile: [SORRY!],  it says `lSafetyData ∙= safetyData1`, which looks different.  We are using _different_ lenses in the two contexts, even though they're equivalent (except that they have different type constraints in the Haskell code, which is about granularity of access capabilities.

Let's discuss at higher bandwidth next time we meet.

haroldcarr: Now I see - you are talking about a USE of `lSafetyData`.
In that case, the branch/location you point to would be better to do the update so it looks like the Haskell code.

mark-moir: Yes, the USE in a future branch :smile:.  Sorry for being unclear and expecting you to be clairvoyant (I still expect it though :smile:).  My comments are regarding _how_ we can make it look like the Haskell code, given that the Haskell code is using a lens that is created by `obmMakeClassy` and we don't have an Agda counterpart for that.

haroldcarr: `obmMakeClassy` is not really the issue. 
That said, let's address this at the use site.

PR #44, opened 2021-06-17
Impl / ImplFake / ImplShared

haroldcarr:

This separates the implementation code into
- `Impl` - the model of the Haskell version
- `ImplFake` - the fake stuff
- `ImplShared` - used by both of the above


cwjnkins (approved):
This looks good to me!

LibraBFT/Impl/Consensus/RoundManager.agda
```diff
@@ -1,82 +1,32 @@
 {- Byzantine Fault Tolerant Consensus Verification in Agda, version 0.9.
 
-   Copyright (c) 2020, 2021, Oracle and/or its affiliates.
```
mark-moir, line 3: We need to keep the years from the beginning of the file's life in the copyright notices, which is from 2020 in this case.

LibraBFT/Concrete/Obligations/PreferredRound.agda
```diff
@@ -3,10 +3,11 @@
...
    Licensed under the Universal Permissive License v 1.0 as shown at https://opensource.oracle.com/licenses/upl
 -}
-open import LibraBFT.Prelude
-open import LibraBFT.Lemmas
 open import LibraBFT.Base.Types
-open import LibraBFT.Impl.Base.Types
+open import LibraBFT.ImplShared.Base.Types
+open import LibraBFT.Lemmas
+
+open import LibraBFT.Prelude
```
mark-moir, line 10: For better or worse, our [style guide])(https://github.com/oracle/bft-consensus-agda/blob/main/STYLE.md#coding-style-conventions), it says to list common imports (like `LibraBFT.Prelude` and `Optics.All`) first, and I think we've stuck to that fairly consistently.  I know your preference (in Haskell) to list imports in alphabetical order, and I share that preference and try to follow it "to the extent possible" (also mentioned in the style guide).  But note that we cannot stick strictly to this rule (some imports depend on previous ones).  You can argue that the style guide is suboptimal, and maybe we'll agree to change it, in which case we should change everywhere to be consistent with it.  I'm in favour of keeping the style guide as-is, and reverting the changes that mix the really common imports in among others, but not "hard over" on this issue, as long as we understand that this is not part of moving towards strictly alpha-order, which we can't.

LibraBFT/Concrete/Intermediate.agda
```diff
@@ -19,7 +19,7 @@
 -- α-ValidVote)
 
 open import LibraBFT.Prelude
-open import LibraBFT.Impl.Base.Types
+open import LibraBFT.ImplShared.Base.Types
```
mark-moir, line 22: I am in two  minds about `ImplShared`.  On one hand, I think we should keep `Impl` stuff in `Impl` and `FakeImpl` can reach into it if needed, because when `FakeImpl` eventually goes away, `ImplShared` won't make sense anymore.  OTOH, I think most of what you've put under `ImplShared` is stuff that really should be somewhere other than where it is, but that's part of a bigger reorg that's not a priority for now.  So, I'm fine with your approach.

mark-moir (changes requested):
Thanks @haroldcarr.  Looks good, though I do have some comments (nothing substantive) that I'd appreciate discussing/addressing before we merge.

LibraBFT/Impl/Consensus/RoundManager.agda
haroldcarr: The `Impl` `RoundManager` is new code this year.
The `ImplFake` `RoundManager`. has `2020, 2021`
So I assume no change is required here?

LibraBFT/Concrete/Obligations/PreferredRound.agda
haroldcarr: Alpha order is better, because:
- Agda does not complain about duplicate imports (unless there is a conflict.
- So alpha makes it easier to see the duplication.

That said, some import depend on earlier imports.  In that case, they should be put below the alpha section with a blank line between.

LibraBFT/Concrete/Intermediate.agda
haroldcarr: OK.  No change for now.   Easy enough to change in the future if need be.

LibraBFT/Concrete/Obligations/PreferredRound.agda
mark-moir: OK, I'm fine if you want to change the style guide along these lines and conform with  it.  I would just use slightly loser language that doesn't require strictly alphabetical followed by non alphabetical, because I sometimes want to list the import and then the one that depends on it right after.  How about alphabetical order, except where there is a dependency that requires a different order, in which case there should be a blank line between any two imports that aren't in alpha order?  Also, I really think we should go with a lighter blue for the bike shed.

mark-moir: Actually, even less specific is better.  Maybe something like "use blank lines to avoid any confusion due to non-alpha order imports".

haroldcarr: OK.   When I update the request, please check the style guide update.

LibraBFT/Impl/Consensus/RoundManager.agda
mark-moir: Ah got it, thanks.

STYLE.md
```diff
@@ -11,7 +11,9 @@ We will maintain style-related conventions here; please try to adhere to them wh
   - Pragmas, if any
   - Imports
      - In general, list all imports in the earliest place possible.  For example, the top of file should include all imports required by the file's modules except those that must be imported within the module due to a dependence on module parameters or a requirement to open public.
-     - To the extent possible, group related imports together in order.  List common imports such as `LibraBFT.Prelude`, `LibraBFT.Lemmas` and `LibraBFT.Base.Types` first.
+     - List imports alphabetically where possible.
+     - For modules that depend on earlier imports, list them right after the import they depend on.
```
mark-moir, line 15: Super pedantic, but an import might depend on multiple previous imports, which makes it impossible to follow this rule.  Also, in some cases it may make less sense to co-locate imports.  I think we can leave that judgment for coders (and code reviewers), rather than trying to dictate this.  I'm happy with the other changes.

mark-moir (changes requested):

STYLE.md
mark-moir: My specific suggestion is to just remove the second bullet.

mark-moir: Errr, I mean the second of the two new ones :smile:.

mark-moir:
LGTM, thanks @haroldcarr!

PR #46, opened 2021-06-19
replace ₋ with _ (underscore)

haroldcarr:
These change are on the b7317d2 chris-process-proposal branch.
I'm out, so if any discussion/changes required, please handle that while I'm gone.


mark-moir:
@haroldcarr LGTM.  We discussed adding a note to `STYLE.md` about all the different meanings of `_` and how this has added yet another.  But I feel reasonably confident this change won't cause much confusion, and it is certainly easier on the eyes and the fingers.  I've changed the target branch to `main`, resolved conflicts (trivial) and am merging now.  Thanks.

PR #43, opened 2021-06-14
Proved PreferredRoundRule on the concrete side

lisandrasilva:
Proved `PreferredRoundRule` on the concrete side adding 2 new implementation obligations for this proof: `PR-ImplObligation₁` and `PR-ImplObligation₂`.  Besides that, we also modified the first implementation obligation of the `VotesOnceRule` to the new one `IncreasingRoundObligation`. We fixed the proof of `VotesOnce` using this new implementation obligation and also used it to prove the `PreferredRoundRule`.

PR #48, opened 2021-06-22
Refactor so we can use different implementation handlers (e.g. fake and real) with the same underlying types

mark-moir:
The main purpose of this PR is to enable us to hook up the under-development implementation model to the concrete properties without breaking or losing our fake implementation and proofs about it.  This further enables proving multiple variants on an implementation.  The main changes are:

* Refactor `Yasm` so that we first instantiate it with the types, and subsequently instantiate it with initialisation information and handlers.
* Tease apart `Concrete` namespace from `Impl*`, so that anything it needs to know about an implementation is provided as a module parameter, not directly from an implementation as it did before.
   * As a result, there is one new proof (`NewVoteRound≢0`) required in `LibraBFT.Concrete.Properties.VotesOnce` and an implementation obligation is provided about honest votes that should enable this proof.
* Move initialisation-related information (mostly postulated for now) into `LibraBFT.Impl.Consensus.Types`, so both `Impl` and `FakeImpl` can share it.
* Establish `ImplementationObligations` for the fake handler (`FakeInitAndHandlers`); the real implementation will need one too, and it can use the same `genesisInfo` and `initRM`, while providing its own handlers.

I had some naming conflict pain and couldn't easily resolve it the "right" way, so had to rename some things to avoid the conflict.  Not ideal but expedient.

I have tried to separate this work into coherent commits (including a couple that are indent-only), but there is no guarantee that each commit builds successfully.

@cwjnkins, this should enable you to establish a handler for the real implementation model, and work towards creating `ImplObligations` for it, analogous to `theImplObligations` in `LibraBFT.ImplFake.Properties`.

@lisandrasilva, assuming we merge this or something like it, it should be a good time to port a version of your proofs in #43.  I would suggest looking at the changes to `LibraBFT.Concrete.Properties.VotesOnce` first, as you will likely have to make similar changes.  P.S., I bet you can't prove `NewVoteRound≢0` 😉

### DIFF
--- a/LibraBFT/Concrete/Properties/VotesOnce.agda
+++ b/LibraBFT/Concrete/Properties/VotesOnce.agda
@@ -3,7 +3,7 @@
    Copyright (c) 2020, 2021, Oracle and/or its affiliates.
    Licensed under the Universal Permissive License v 1.0 as shown at https://opensource.oracle.com/licenses/upl
 -}
-{-# OPTIONS --allow-unsolved-metas #-}
+
 open import LibraBFT.Base.KVMap
 open import LibraBFT.Base.PKCS
 open import LibraBFT.Concrete.System
--- a/LibraBFT/Concrete/Properties/VotesOnce.agda
+++ b/LibraBFT/Concrete/Properties/VotesOnce.agda
@@ -3,22 +3,20 @@
    Copyright (c) 2020, 2021, Oracle and/or its affiliates.
    Licensed under the Universal Permissive License v 1.0 as shown at https://opensource.oracle.com/licenses/upl
 -}
-
+{-# OPTIONS --allow-unsolved-metas #-}
 open import LibraBFT.Base.KVMap
 open import LibraBFT.Base.PKCS
 open import LibraBFT.Concrete.System
 open import LibraBFT.Concrete.System.Parameters
-open import LibraBFT.ImplFake.Handle
-open import LibraBFT.ImplFake.Handle.Properties
 open import LibraBFT.ImplShared.Base.Types
 open import LibraBFT.ImplShared.Consensus.Types
 open import LibraBFT.ImplShared.Util.Crypto
 open import LibraBFT.Lemmas
 open import LibraBFT.Prelude
+open import LibraBFT.Yasm.Base
 open import Optics.All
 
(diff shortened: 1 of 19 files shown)


LibraBFT/Concrete/Properties/VotesOnce.agda
```diff
@@ -3,22 +3,20 @@
    Copyright (c) 2020, 2021, Oracle and/or its affiliates.
    Licensed under the Universal Permissive License v 1.0 as shown at https://opensource.oracle.com/licenses/upl
 -}
-
+{-# OPTIONS --allow-unsolved-metas #-}
```
cwjnkins, line 6: I will push a commit to remove this (there are no unsolved metas here).

cwjnkins:
From what I can tell, this looks good! The best test though is for me to merge this into my current work and see if I can write stubs for the proofs of the implementation obligations for the more realistic implementation, so I will do that now.

cwjnkins:
@mark-moir This is perhaps only partially related, but I think that it would make sense for this PR to also move the  code in `LibraBFT/ImplFake/Handle.agda` responsible for converting `List Output` to `List (Action NetworkMsg)`, since all types mentioned are shared between both the fake and real implementation handler.

(Context: to connect the contracts for the peer handler to the system state, I am finding we will need some lemmas relating the outputs of e.g. `processProposalMsgM` to the list of actions produced from this using `outputsToActions`).

cwjnkins:
Oh, I see now that since this PR is on a branch which still has `MetaEpochIndep.agda`, moving `outputsToActions` near the definition of `Output` in this PR does not make sense -- better to wait until we remove `MetaEpochIndep.agda` from the main branch.

cwjnkins (approved):
This looks good to me!

cwjnkins:
Oh, I see now that since this PR is on a branch which still has `MetaEpochIndep.agda`, moving `outputsToActions` near the definition of `Output` in this PR does not make sense -- better to wait until we remove `MetaEpochIndep.agda` from the main branch.

mark-moir:
Great, thanks @cwjnkins.  I asked @haroldcarr to review only to keep him in the loop, but he's out, so I'll merge now.

PR #50, opened 2021-06-23
Refactor `Output` and friends to `ImplShared`, start modeling network handlers

mark-moir:
This PR refactors some of the types and functions defined with the `FakeImpl` handler into `ImplShared` to enable them to be used also by the real implementation model.  It further begins modeling the actual network handlers so we can see where to connect the pieces.


cwjnkins (approved):
This looks good to me!

PR #52, opened 2021-06-23
LibraBFT: ImplShared: Consensus: Types: amend NO-DEPENDENT-LENSES comment

cwjnkins:
Just a small amendment to this comment. One of the examples, (`bs <- gets rmGetBlockStore`) would not have been well-typed.


mark-moir:
Good catch, thanks @cwjnkins.  Merging now.

PR #55, opened 2021-06-24
A couple of proofs of `IsValidProposalM.contract`, room for improvement!

mark-moir:
There are two versions of the proof in two commits.  I think/hope I am missing something to make this more streamlined, so I am sharing for discussion and improvement.  I'd like to improve and document this more before merging.


cwjnkins:
@mark-moir I've pushed my solution, we can discuss later today.

PR #56, opened 2021-06-25
Chris monad and eq instances

cwjnkins:
I have added some definitions to allow do-notation for all sorts of monads, and defined instances for `Either` and (the old) `RWST`. To demonstrate the need, I have implemented `processProposal` in `LibraBFT.Impl.Consensus.Network`.

Also, I have gone ahead and added an `Eq` record, opened with Agda's instance argument syntax, to allow writing `_==_` for equality tests. If we change the implementation such that `NetworkMsg` contains the sender of the message, this could be used to complete the definition of `processProposal` that follows the Haskell prototype closely (the code that would do this is currently commented out).


mark-moir:
I believe this will be superseded by coming merge requests that come from the discussion in #58, so I'm closing it.

PR #59, opened 2021-06-28
Overhaul the `RWST` monad and add support for reasoning with Dijkstra's weakest precondition calculus.

cwjnkins:
This PR also:

- adds a `Monad` record opened with Agda's instance arguments, to simulate Haskell's type classes (this allows us to use do-notation for `Either` as well as `LBFT`)
- moves the convenient syntax for do-notation with `RWST` to its own file. This module is imported publicly by `LibraBFT/ImplShared/Util/Util.agda`, so many occurrences of `open RWST-do` have been removed from the implementation models.
- replaces some `if` and `case` statements with the branching operators of the new `RWST` datatype
- removes `ErrLog` and replaces all remaining occurrences with `FakeErr`

### DIFF
--- a/LibraBFT/Impl/Consensus/BlockStorage/SyncManager.agda
+++ b/LibraBFT/Impl/Consensus/BlockStorage/SyncManager.agda
@@ -19,7 +19,5 @@ open import Optics.All
 
 module LibraBFT.Impl.Consensus.BlockStorage.SyncManager where
 
-open RWST-do
-
 postulate
   insertQuorumCertM : QuorumCert → BlockRetriever → LBFT (Either ErrLog Unit)
--- a/LibraBFT/Impl/Consensus/BlockStorage/SyncManager.agda
+++ b/LibraBFT/Impl/Consensus/BlockStorage/SyncManager.agda
@@ -20,4 +20,4 @@ open import Optics.All
 module LibraBFT.Impl.Consensus.BlockStorage.SyncManager where
 
 postulate
-  insertQuorumCertM : QuorumCert → BlockRetriever → LBFT (Either ErrLog Unit)
+  insertQuorumCertM : QuorumCert → BlockRetriever → LBFT (Either FakeErr Unit)
--- a/LibraBFT/Impl/Consensus/RoundManager.agda
+++ b/LibraBFT/Impl/Consensus/RoundManager.agda
@@ -30,8 +30,6 @@ open import LibraBFT.Abstract.Types.EpochConfig UID NodeId
 
 module LibraBFT.Impl.Consensus.RoundManager where
 
-open RWST-do
-
 ------------------------------------------------------------------------------
 
 processCommitM : LedgerInfoWithSignatures → LBFT (List ExecutedBlock)
--- a/LibraBFT/Impl/Consensus/RoundManager.agda
+++ b/LibraBFT/Impl/Consensus/RoundManager.agda
@@ -54,23 +54,19 @@ module processProposalMsgM (now : Instant) (pm : ProposalMsg) where
   step₂ : Either FakeErr Bool → LBFT Unit
 
   step₀ =
-    case pm ^∙ pmProposer of λ where
+    caseMM pm ^∙ pmProposer of λ where
       nothing → logInfo -- log: info: proposal with no author
       (just pAuthor) → step₁ pAuthor
   step₁ pAuthor =
         ensureRoundAndSyncUpM now (pm ^∙ pmProposal ∙ bRound) (pm ^∙ pmSyncInfo)
                               pAuthor true >>= step₂
-  step₂ r =
-        -- IMPL-DIFF: We use `ifM` to test whether the round of the proposal is
-        -- current, to take advantage of the obligations `RWST-weakestPre` generates.
-        case r of λ where
-        (Left _) → logErr -- log: error: <propagate error>
-        (Right pmCurrent) →
-          if pmCurrent
-            then processProposalM (pm ^∙ pmProposal)
-            else do
-              currentRound ← use (lRoundState ∙ rsCurrentRound)
-              logInfo  -- log: info: dropping proposal for old round
+  step₂ =
+        λ where
+          (Left e)      → logErr -- log: error: <propagate error>
+          (Right true)  → processProposalM (pm ^∙ pmProposal)
+          (Right false) → do
+            currentRound ← use (lRoundState ∙ rsCurrentRound)
+            logInfo              -- log: info: dropping proposal for old round
 
 processProposalMsgM = processProposalMsgM.step₀
 
@@ -90,14 +86,14 @@ module ensureRoundAndSyncUpM
 
   step₀ = do
     currentRound ← use (lRoundState ∙ rsCurrentRound)
-    if ⌊ messageRound <? currentRound ⌋
+    ifM messageRound <? currentRound
       then ok false
       else step₁
   step₁ =
         syncUpM now syncInfo author helpRemote ∙?∙ λ _ → step₂
   step₂ = do
           currentRound' ← use (lRoundState ∙ rsCurrentRound)
-          if not ⌊ messageRound ≟ℕ currentRound' ⌋
+          ifM not ⌊ messageRound ≟ℕ currentRound' ⌋
             then bail fakeErr -- error: after sync, round does not match local
             else ok true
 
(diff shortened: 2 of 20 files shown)


LibraBFT/ImplShared/Util/RWST.agda
```diff
@@ -5,175 +5,236 @@
...
-open import Optics.All
 
--- This module defines syntax and functionality modeling an RWST monad,
--- which we use to define an implementation.
+-- This module defines functionality modeling an RWST monad, which we use to
+-- define an implementation, and functionality for proving properties about
+-- programs written using this RWST monad. The main definitions are:
+-- 1. RWST, a datatype for the ASTs of stateful programs that read from an
+--    environment and produce output.
+--    This datatype includes constructors for branching code, to aide in the
```
mark-moir, line 14: s/aide/aid/

LibraBFT/ImplShared/Util/RWST.agda
```diff
@@ -5,175 +5,236 @@
...
+                → (A → RWST Ev Wr St (Either C B))                → RWST Ev Wr St (Either C B)
+  RWST-maybe  : ∀ {A B} → Maybe B
+                → (RWST Ev Wr St A) → (B → RWST Ev Wr St A)       → RWST Ev Wr St A
+
+private
+  variable
+    Ev Wr St : Set
+    A B C    : Set
+
+-- To execute an RWST program, you prove an environment and prestate. This
```
mark-moir, line 59: s/prove/provide/ ?

LibraBFT/ImplShared/Util/RWST.agda
```diff
@@ -5,175 +5,236 @@
...
+RWST-weakestPre-ebindPost ev f Post (Right r) post outs =
+  ∀ c → c ≡ r → RWST-weakestPre (f c) (RWST-Post++ Post outs) ev post
+
+RWST-weakestPre-bindPost ev f Post x post outs =
+  ∀ r → r ≡ x → RWST-weakestPre (f r) (RWST-Post++ Post outs) ev post
+
+-- For every RWST computation `m`, `RWST-Contract m` is the type of proofs that,
+-- for all post conditions `P`, starting environments `ev` and prestates `pre`,
+-- to prove that `P` holds after running `m` in `ev` and `pre`, it suffices to
+-- provide the weakest precondition for `P` with respect to `m`, `ev`, and `pre`.
```
mark-moir, line 160: s/provide/prove/ ?

LibraBFT/ImplShared/Util/RWST.agda
```diff
@@ -5,175 +5,236 @@
...
+RWST-Contract : (m : RWST Ev Wr St A) → Set₁
+RWST-Contract{Ev}{Wr}{St}{A} m =
+  (P : RWST-Post Wr St A)
+  → (ev : Ev) (pre : St) → RWST-weakestPre m P ev pre
+  → let (x , post , outs) = RWST-run m ev pre in
+    P x post outs
+
+-- This proves `RWST-Contract m` for arbitrary `m`.
```
mark-moir, line 168: It would be helpful to add something like: This proves that our definition of `RWST-weakestPre` provides a _sufficient_ precondition for the post condition to hold.  It does _not_ prove that it is the weakest such precondition (even though it's true); this is not important for our purposes.

mark-moir (approved):
Looks great @cwjnkins, thanks!  I've just made a few minor comments on spelling and comments.  Feel free to address as you see fit and merge.

cwjnkins:
@mark-moir I have made the requested changes. I will wait until later today (after we meet) to merge this in.

LibraBFT/Impl/Consensus/RoundManager.agda
```diff
@@ -56,23 +54,19 @@ module processProposalMsgM (now : Instant) (pm : ProposalMsg) where
   step₂ : Either FakeErr Bool → LBFT Unit
 
   step₀ =
-    case pm ^∙ pmProposer of λ where
+    caseMM pm ^∙ pmProposer of λ where
```
haroldcarr, line 57: What does `MM` mean?
Why isn't `case` good enough here?

LibraBFT/Impl/Consensus/RoundManager.agda
```diff
@@ -92,14 +86,14 @@ module ensureRoundAndSyncUpM
 
   step₀ = do
     currentRound ← use (lRoundState ∙ rsCurrentRound)
-    if ⌊ messageRound <? currentRound ⌋
+    ifM messageRound <? currentRound
```
haroldcarr, line 89: I'm definitely confused.

In the haskell code, the `M` suffix was generally used to mean functions in the `LBFT` monad.

This particular `ifM` seems that it is turning the decidable test into a boolean.   That has nothing to do with a monad nor a `LBFT` monad.

LibraBFT/Impl/Consensus/RoundManager.agda
```diff
@@ -121,16 +115,15 @@ module ProcessProposalM (proposal : Block) where
...
+  step₃ : Vote → SyncInfo → LBFT Unit
 
   step₀ = do
     s ← get  -- IMPL-DIFF: see comment NO-DEPENDENT-LENSES
     let bs = rmGetBlockStore s
     vp ← ProposerElection.isValidProposalM proposal
     step₁ {s} bs vp
   step₁ bs vp =
-    grd‖ is-nothing (proposal ^∙ bAuthor) ≔
+    ifM‖ is-nothing (proposal ^∙ bAuthor) ≔
```
haroldcarr, line 126: If I am right about my earlier comment, then this should be `if||`

haroldcarr:
before I continue reviewing, I am going to wait for responses to my existing comments since I seem to be missing some context

haroldcarr:
Approved - pending answer to why `FakeErr` rather than `ErrLog`

haroldcarr (approved):

LibraBFT/Impl/Consensus/BlockStorage/SyncManager.agda
```diff
@@ -19,7 +19,5 @@ open import Optics.All
 
 module LibraBFT.Impl.Consensus.BlockStorage.SyncManager where
 
-open RWST-do
-
 postulate
-  insertQuorumCertM : QuorumCert → BlockRetriever → LBFT (Either ErrLog Unit)
+  insertQuorumCertM : QuorumCert → BlockRetriever → LBFT (Either FakeErr Unit)
```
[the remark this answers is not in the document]
cwjnkins: This was the name @mark-moir introduced. I am also not sure why it is `FakeErr`.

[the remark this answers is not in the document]
mark-moir: I'm sure we discussed this, but for the record...

The point of `FakeErr` is to serve as a placeholder in our model (Agda implementation of the Haskell code) for where the Haskell code logs actual errors, which have a bunch of constructors, most of which take `[Text]` arguments.  We don't want to spend time modeling these errors in detail at this stage (or maybe ever, at least in cases we can prove don't ever happen, e.g., internal invariant checks).  We were initially not modeling the errors at all, but this would affect the "shape" of proofs, so "we" decided to have a placeholder with no structure.  HTH.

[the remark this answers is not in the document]
mark-moir: Perhaps a better way would be to keep `ErrLog`, but just model it with a single constructor `FakeErr`.  That way we could incrementally add real constructors later, perhaps eventually eliminating the`FakeErr` constructor, or spell it differently such as `ImpossibleError`, and then prove that nothing ever emits `ImpossibleError`.  Thoughts?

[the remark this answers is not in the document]
cwjnkins: This was the name @mark-moir introduced. I am also not sure why it is `FakeErr`.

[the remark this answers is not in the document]
cwjnkins: I don't think there is a need to prove that `ImpossibleError` is never emitted when all the errors have been implemented -- we can just remove that constructor. I think having a datatype for `ErrLog` and *postulating* `fakeErr` (so that no proofs ever match against it) might be a good way to go.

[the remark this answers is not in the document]
mark-moir: My point was that maybe there's not much point modeling the details of errors that we can prove are never emitted, but yes, we could  in principle get to a fully detailed model in time; the maybe-temporary existence of `FakeError` or `ImpossibleError` will enable us to get there incrementally.  Regarding postulating `fakeErr : ErrLog`, I thought that was how I started and you didn't like it :smile:.  Maybe I'm confused though.  In any case, it seems reasonable to me, and this conversation has made me want to make our type signatures look identical again :smile:.

mark-moir:
@cwjnkins AFAIAC, this is ready to merge.  Merge at will!  Thanks.

PR #61, opened 2021-07-01
Logging in, and proofs about, `LibraBFT.Impl.Consensus.SafetyRules.SafetyRules`

cwjnkins:
This PR adds info logging to the functions in `SafetyRules`, and re-does the proofs from `chris-dijkstra-monad` to account for these changes.

### DIFF
--- a/LibraBFT/Impl/Consensus/RoundManager/PropertyDefs.agda
+++ b/LibraBFT/Impl/Consensus/RoundManager/PropertyDefs.agda
@@ -25,8 +25,8 @@ NoOutputs outs = outs ≡ []
 NoMsgOuts : List Output → Set
 NoMsgOuts outs = List-filter isOutputMsg? outs ≡ []
 
--- Invariants that all peer handlers must observe.
-record Invariant₂ (pre post : RoundManager) : Set where
+record NoEpochChange (pre post : RoundManager) : Set where
+  constructor mkNoEpochChange
   field
     es≡₁ : (_rmEC pre) ≡L (_rmEC post) at rmEpoch
     es≡₂ : pre ≡L post at lSafetyData ∙ sdEpoch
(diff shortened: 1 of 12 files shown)


LibraBFT/Impl/Consensus/SafetyRules/Properties/SafetyRules.agda
```diff
@@ -0,0 +1,717 @@
...
+    step₃-contract
+      : ∀ pre safetyData1 voteData author ledgerInfo
+        → LBFT-weakestPre
+            (step₃ safetyData1 voteData author ledgerInfo)
+            Contract pre
+    step₃-contract pre safetyData1 voteData author ledgerInfo
+    {-
+    The proof can be as simple as this:
+
+       = λ _ _ _ _ _ _ → refl
```
cwjnkins, line 412: Making a note to myself that the essay still needs fixing.

mark-moir: More generally, I have not reviewed all of the proofs in details (however, Agda has :smile:).  I don't want to hold up the merge to do so.  I'd like to go over especially the proofs labeled as "tutorial" and see how far they are from being understandable by mortals such as myself.  Then I can make or suggest changes later if needed.

LibraBFT/Impl/Consensus/SafetyRules/SafetyRules.agda
```diff
@@ -145,15 +159,15 @@ module constructAndSignVoteM-continue2 (voteProposal : VoteProposal) (validatorS
   step₁ safetyData1 = do
     lSafetyData ∙= safetyData1  -- TODO-1: resolve discussion about pssSafetyData vs lSafetyData
     extensionCheckM voteProposal ∙?∙ (step₂ safetyData1)
-
```
mark-moir, line 148: You've fairly consistently left no blank lines between steps.  (There is still at least one, between steps 0 and 1 in `constructAndSignVoteM-continue2`.)  I'd actually prefer to consistently *include* the blank lines, as I think it makes it easier to compare to the Haskell, as we need to visually "elide" the lines that introduce the next step and the arguments from the previous one.  If you feel otherwise, that's OK, I'll try to get used to it.

mark-moir (approved):
@cwjnkins great stuff, thanks!  I have made a number of review comments, but most are superficial things about comments and style.  Please address them as you see fit, and merge when you're happy (or let me know if you want to discuss anything first).

LibraBFT/Impl/Consensus/SafetyRules/Properties/SafetyRules.agda
cwjnkins: @mark-moir The fix I had in mind for the essay was only to remove two underscores. I am hoping to keep the proof in the private module `Tutorial` completely separate from the actual proof we use, because these proofs are going to undergo changes as more properties are tracked. There is certainly value in breaking up the larger proofs (that will be used to prove e.g. the votes once property) into smaller pieces, but I think the extended essay should be for something stand-alone since a considerable amount of effort goes into updating that essay if the model or property is modified.

LibraBFT/ImplShared/Interface/Output.agda
```diff
@@ -31,12 +31,26 @@ module LibraBFT.ImplShared.Interface.Output where
   IsSendVote : Output → Set
   IsSendVote out = ∃₂ λ mv pid → out ≡ SendVote mv pid
 
+  IsBroadcastProposal : Output → Set
```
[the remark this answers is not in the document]
cwjnkins: At the moment, I don't see a major advantage for my purposes one way or the other. If its more convenient for your efforts to have them phrased this way, then lets do it this way.

LibraBFT/Impl/Consensus/SafetyRules/SafetyRules.agda
cwjnkins: I will add those extra lines in.

LibraBFT/Impl/Consensus/RoundManager/PropertyDefs.agda
```diff
@@ -0,0 +1,74 @@
...
+NoOutputs : List Output → Set
+NoOutputs outs = outs ≡ []
+
+NoMsgOuts : List Output → Set
+NoMsgOuts outs = List-filter isOutputMsg? outs ≡ []
+
+-- Invariants that all peer handlers must observe.
+record Invariant₂ (pre post : RoundManager) : Set where
+  field
+    es≡₁ : (_rmEC pre) ≡L (_rmEC post) at rmEpoch
```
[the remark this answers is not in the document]
cwjnkins: This is a good point. I have renamed it to `NoEpochChange`, so there is no longer a suggestion that the lack of epoch change is an invariant.

PR #65, opened 2021-07-01
fix top level indent of 3 new Impl files

haroldcarr:
We have agreed that ALL new files will not indent the top level module, e.g.,

```
module foo where

stuff (not indented)

more stuff
```

This commit "fixes" 3 new files that were still indenting.


cwjnkins (approved):
Looks good to me.

PR #66, opened 2021-07-02
finish parts of RoundManager processVoteM

haroldcarr:
- `RoundManager.processVoteM`
  - replaced a stub with a real call to `ProposerElection.isValidProposer`
    - this required implementing `rmPgAuthor` in `ImplShared.Consensus.Types`
      - NOTE: in Haskell this is a getter only lens `pgAuthor` whereas in Agda I have made it a function.
      - Do we have a way to define getter only?
  - replaced a stub with a real call to `BlockStore.getQuorumCertForBlock`
    - the stub was written before `rmGetBlockStore` existed so now can be replaced
- `RoundManager.addVoteM`
  - use `rmGetBlockStore`
  - this required adding various fields, lens and functions to timeout data
  - CHRIS: there is a note in the code pointing out a type checking problem I could use help on.

### DIFF
--- a/LibraBFT/Impl/Consensus/RoundManager.agda
+++ b/LibraBFT/Impl/Consensus/RoundManager.agda
@@ -225,13 +225,10 @@ processVoteM now vote =
 addVoteM now vote = do
   s ← get -- IMPL-DIFF: see comment NO-DEPENDENT-LENSES
   let bs = rmGetBlockStore s
-  -- CHRIS: could you uncomment this (and remove the continue below)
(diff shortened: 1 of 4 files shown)


cwjnkins:
We do not have a way yet to define getter-only lenses (though we could implement this). My thinking is we should go ahead an make this an ordinary lens. The Haskell type system will enforce that this lens cannot be used to set that field -- since the Agda code follows the Haskell closely, the corresponding Agda lens won't be used to set the field either.

cwjnkins (approved):

PR #68, opened 2021-07-07
Refactor `RoundManager` and lens support

mark-moir:
**For discussion only, do not merge**

An experimental approach to allowing us to use all the lenses in the Agda that we use in the Haskell code.  The key idea is to keep *only* the meta data (particularly regarding validity of stored QCs) in the epoch-dependent types.  The approach is to define a type `RMLens` that carries evidence that we can compute a new `RoundManager` (including that the epoch is correct and the meta data needed) if we use the lens to update the state.  This will be easy to show for many lenses that don't touch information from which `EpochConfig` is computed and that don't modify the stored `QC`s.  Other cases may be more difficult.  I have added "pseudocomposition", enabling composition of an `RMLens` and an ordinary lens.  I have updated almost all of the code to look more like the Haskell, and it appears promising.  

However, there are some issues still, including:
* Annoying need to create aliases for lots of lenses  because Agda can't parse types defined in terms of autogenerated Lenses (not that big a deal really, just annoying).
* The reality (whether we make these changes or not) that updates to fields about which we want to store metadata might require application-specific proofs that we can either reuse the old metadata, or construct new metadata presents some challenges.
* I have made some mistakes, in particular misunderstanding the nature of the `l*` lenses (e.g., `lSafetyData`), and handling them in the wrong way.  I think I know what to do, I just haven't got to it yet.

Further feedback welcome.


LibraBFT/ImplShared/Consensus/Types.agda
```diff
@@ -162,6 +202,66 @@ module LibraBFT.ImplShared.Consensus.Types where
...
+    g : RoundManagerEC → Bool
+    g rmec = _rmSyncOnly rmec
+
+    s : RoundManagerEC → Bool → RoundManagerEC
+    s rmec b = record rmec { _rmSyncOnly = b}
+
+  instance
+     -- Note that this is an autogenerated lens that doesn't come anywhere near anything to do with
+     -- epoch change, so is a good test example, though we can always define the top-level lenses
+     -- manually if this doesn't work.  Maybe that's where I should start.
```
cwjnkins, line 231: On this point, I think that a pseudo-composition operator is needed for`GoodLens`. Specifically, composing a `GoodLens` with a (compatible) `Lens` produces a new `GoodLens`.

LibraBFT/ImplShared/Consensus/Types.agda
```diff
@@ -162,6 +202,66 @@ module LibraBFT.ImplShared.Consensus.Types where
   α-EC-RM : RoundManager → EpochConfig
   α-EC-RM rm = α-EC ((_rmEC rm) , (_rmEC-correct rm))
 
+
+  record GoodLens {A} (l : Lens RoundManagerEC A) : Set where
```
cwjnkins, line 206: How about something like`RMLens` for the name?

cwjnkins (changes requested):
@mark-moir  Over all this looks good, but I would like to see some of the functions in `RoundManager.agda` and `SafetyRules.agda` changed to use `GoodLens`s (in particular, the parts that need to fetch the block store) to make the proposal more concrete.

LibraBFT/ImplShared/Util/Util.agda
```diff
@@ -23,6 +23,10 @@ module LibraBFT.ImplShared.Util.Util where
   LBFT : Set → Set₁
   LBFT = RWST Unit Output RoundManager
 
+  LBFT-Monad : Monad LBFT
```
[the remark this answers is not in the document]
mark-moir: Yep, this was a result of some confusion on my part.  Gone.

LibraBFT/ImplShared/Consensus/Types.agda
mark-moir: Sounds good, thanks.  Done.

mark-moir:
@cwjnkins I've pushed some more changes and updated the description.  It's a bit messy and there are some issues as discussed in the description.  But still it might be worth having another quick look and letting me know your thoughts.  Thanks.

mark-moir:
@cwjnkins @haroldcarr On further reflection, I think it was a mistake to separate the `BlockStore` and associated metadata.  We can get the same benefits using something like `RMLens` without complicating things by also rearranging the data and metadata.

cwjnkins:
> On further reflection, I think it was a mistake to separate the `BlockStore` and associated metadata. We can get the same benefits using something like `RMLens` without complicating things by also rearranging the data and metadata.

@mark-moir How so? If `BlockStore` depends on the epoch config, I don't see how we can write a lens to access it from `RoundManager` without dependent lenses and a dependent `RWST` monad.

mark-moir:
> > On further reflection, I think it was a mistake to separate the `BlockStore` and associated metadata. We can get the same benefits using something like `RMLens` without complicating things by also rearranging the data and metadata.
> 
> @mark-moir How so? If `BlockStore` depends on the epoch config, I don't see how we can write a lens to access it from `RoundManager` without dependent lenses and a dependent `RWST` monad.

Maybe I'm being naive, but I think the `RMLens` approach will allow us to have almost all the lenses we want.  The key is that the the variants in `Utils.Utils` know about our types, so we can convince them in almost all cases to work.  The one case that I think will remain is that we can't have lenses for any field that we can't prove won't change the `EpochConfig`.  In the Haskell code, most/all of those are `Getter`s only, so I'm thinking about whether we could have the same for use with the variants in `Util.Util`.

mark-moir:
Closing this as it is superseded by #71 after we decided to keep proof metadata in proofs, not in `RoundManager`; our original approach caused too much pain for not enough benefit.

PR #69, opened 2021-07-07
Proof that the two Epoch fields of `RoundManager` stay the same, and misc.

cwjnkins:
This PR adds `LibraBFT.Impl.Handle.Properties`, and the main proof there is of the following system invariant: the two different `Epoch` fields in `RoundManager` remain the same. This relies on a postulated contract for `handleProposal` which is based on existing proofs for related `RoundManager` functions for processing a proposal.

This PR also adds:
- Some proof utilities for `RWST` and `LBFT`
- a `ToBool`-polymorphic `not`
- an `Eq` type class
- a new lemma about `List-filter`
- changes to the initial `RoundManager` for all peers (more fields populated)

### DIFF
--- a/LibraBFT/Impl/Consensus/Liveness/ProposerElection.agda
+++ b/LibraBFT/Impl/Consensus/Liveness/ProposerElection.agda
@@ -30,4 +30,4 @@ isValidProposalM b =
 
 isValidProposerM a r = isValidProposer <$> use lProposerElection <*> pure a <*> pure r
 
-isValidProposer pe a r = toBool $ getValidProposer pe r == a
+isValidProposer pe a r = getValidProposer pe r == a
--- a/LibraBFT/Impl/Consensus/Liveness/ProposerElection.agda
+++ b/LibraBFT/Impl/Consensus/Liveness/ProposerElection.agda
@@ -30,4 +30,4 @@ isValidProposalM b =
 
 isValidProposerM a r = isValidProposer <$> use lProposerElection <*> pure a <*> pure r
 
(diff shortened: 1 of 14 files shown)


LibraBFT/Impl/Consensus/Liveness/ProposerElection.agda
```diff
@@ -30,4 +30,4 @@ isValidProposalM b =
 
 isValidProposerM a r = isValidProposer <$> use lProposerElection <*> pure a <*> pure r
 
-isValidProposer pe a r = ⌊ getValidProposer pe r ≟ℕ a ⌋
+isValidProposer pe a r = toBool $ getValidProposer pe r == a
```
mark-moir, line 33: Very minor nit, but the `toBool $` reveals that the `==` isn't quite equivalent to `==` in Haskell.  Given that, is this any better than  keeping `⌊ ... ⌋`?  I'm not saying you should make it so the code can look identical, unless you feel the urge and can do it without too much pain.

mark-moir:
Looks great @cwjnkins, approved to merge with or without addressing my one minor comment.  Thanks.

LibraBFT/Impl/Consensus/Liveness/ProposerElection.agda
cwjnkins: The most lightweight change would be to change the type of `isValidProposer`, so I'll do that.

cwjnkins: Scratch my last comment -- I've changed it so that `_==_` returns `Bool`, as it does in Haskell.

PR #71, opened 2021-07-09
Eliminate metadata from state to solve issues with lack of dependent lenses

mark-moir:
The pull request supersedes #68 because we have decided to give up on tracking proof metadata in the state, and instead maintain it in invariants we prove.  This eliminates some painful issues with the lack of dependent lenses, preventing us from faithfully modeling the code we are verifying.  Now:

* `Consensus.Types` is unaware of `EpochConfig`; definitions related to `EpochConfig` are now in `Types.EpochDep`, which now has an inner module `WithEC` that is parameterised by an `EpochConfig`
* We can have all the lenses, and the previous workarounds have been eliminated, which eliminates a number of differences between the Haskell and Agda codes.
* All the proofs are fixed up; only a couple required any thinking; these were due to the proof obligations changing slightly when we switch from patterns like:
```
st <- get
let x = _someSpecialFunction st
...
```
to
```
x <- use aRegularLens
```

### DIFF
--- a/LibraBFT/Concrete/Obligations.agda
+++ b/LibraBFT/Concrete/Obligations.agda
@@ -4,10 +4,14 @@
    Licensed under the Universal Permissive License v 1.0 as shown at https://opensource.oracle.com/licenses/upl
 -}
 
+open import LibraBFT.ImplShared.Base.Types
+
+open import LibraBFT.Abstract.Types.EpochConfig UID NodeId
 open import LibraBFT.Base.PKCS
 open import LibraBFT.Concrete.System.Parameters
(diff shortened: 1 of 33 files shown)


cwjnkins (commented):
This looks good! Before merging however, I will add an example property and contract for keeping track that every quorum in the block store is valid.

cwjnkins:
@mark-moir I have sketched in the latest commit a proposal for how the properties that were previously tracked in `RoundManager` itself can be proven extrinsically.

mark-moir:
Thanks @cwjnkins, this looks great!

PR #77, opened 2021-07-14
SafetyRules.extensionCheck

haroldcarr:
@cwjnkins

at the appropriate time, please switch the code to use `extensionCheck` (instead of `extensionCheckM`, which should then be deleted).


mark-moir:
@haroldcarr, I find this a bit confusing.  Is the Haskell code also going to change to use a pure variant of `extensionCheck`?  It does not seem to at the moment.

haroldcarr:
The Haskell code DOES use the pure version.   You need to update your copy of the Haskell branch.

mark-moir:
> The Haskell code DOES use the pure version. You need to update your copy of the Haskell branch.

Yep, you're right sorry.  I did fetch, I just didn't Fetch :smile:.  LGTM.  Thanks.

PR #78, opened 2021-07-14
Make `verifyAndUpdatePreferredRound` use `compare`, revise and prove the Contract for it

mark-moir:
This PR changes the implementation of `verifyAndUpdatePreferredRound` to use `compare`, consistent with the Haskell code, establishes a Contract for it that actually holds (the original one assumed no `Output`s but the implementation does logging), and proves it.  It also deletes a file that should not have been checked in.


LibraBFT/Impl/Consensus/SafetyRules/Properties/SafetyRules.agda
```diff
@@ -43,18 +43,55 @@ module verifyAndUpdatePreferredRoundMSpec (quorumCert : QuorumCert) (safetyData
...
-        → ((1cr<pr : C₁) → P (inj₁ fakeErr) pre [])
-        → ((1cr≥pr : ¬ C₁)
-           → ((2cr>pr : C₂) → P (inj₂ safetyData') pre [])
-              × ((2cr<pr : C₃) → P (inj₂ safetyData) pre [])
-              × ((2cr=pr : C₄) → P (inj₂ safetyData) pre []))
-        → LBFT-weakestPre (verifyAndUpdatePreferredRoundM quorumCert safetyData) P pre
+  module _ (pre : RoundManager) where
+    record setPR (sd : SafetyData) : Set where
+      field
+        eff   : sd ≡ safetyData'
```
cwjnkins, line 52: This won't be needed any time soon I think, but in principle we should be tracking the condition under which we update the preferred round (i.e., `twoChainRound > preferredRound`) -- otherwise how can we prove e.g. that the preferred round is nondecreasing? Maybe leave a TODO here for this and the corresponding condition (`twoChainRound <= preferredRound`) for `noChanges`.

cwjnkins (changes requested):
One issue with the contract for `verifyAndUpdatePreferredRoundM`, but it could be left as a comment for a future TODO.

LibraBFT/Impl/Consensus/SafetyRules/Properties/SafetyRules.agda
mark-moir: Thanks @cwjnkins, good catch, especially because addressing it revealed that I had proved it as if we update when `twoChainRound = preferredRound`, which isn't true.  FIXED.

PR #82, opened 2021-07-23
Fully connect `handle` in `InputOutputHandlers` to `RoundManager`

cwjnkins:
This PR implements `processProposal` and `processVote` in `LibraBFT.Impl.Consensus.Network.agda` and removes the remaining hole in `LibraBFT.Impl.IO.OBM.InputOutputHandlers.agda`.

### DIFF
--- a/LibraBFT/Impl/Consensus/Network.agda
+++ b/LibraBFT/Impl/Consensus/Network.agda
@@ -22,7 +22,7 @@ processProposal {- peerId -} proposal myEpoch vv =
     (Left e) → Left (Left e)
     (Right unit) →
       grd‖ proposal ^∙ pmProposal ∙ bEpoch == myEpoch ≔
-           return unit
+           pure unit
          -- TODO : push this onto a queue if epoch is in future (is this still relevant?)
          ‖ proposal ^∙ pmProposal ∙ bEpoch == myEpoch + 1 ≔
            Left (Right fakeInfo) -- proposal in new epoch arrived before my epoch change
--- a/LibraBFT/Impl/Consensus/Network.agda
+++ b/LibraBFT/Impl/Consensus/Network.agda
@@ -9,11 +9,28 @@ open import LibraBFT.ImplShared.Base.Types
 open import LibraBFT.ImplShared.Consensus.Types
 open import LibraBFT.ImplShared.NetworkMsg
 open import LibraBFT.ImplShared.Util.Util
+open import LibraBFT.Impl.Consensus.ConsensusTypes.ProposalMsg as ProposalMsg
 open import LibraBFT.Prelude
 open import Optics.All
 
 module LibraBFT.Impl.Consensus.Network where
 
 postulate  -- TODO-1: implement this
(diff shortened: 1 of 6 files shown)


LibraBFT/Impl/Consensus/Network.agda
```diff
@@ -9,10 +9,53 @@ open import LibraBFT.ImplShared.Base.Types
...
 
 module LibraBFT.Impl.Consensus.Network where
 
-postulate  -- TODO-1: implement this
-  processProposal : {- NodeId → -} ProposalMsg → Epoch → ValidatorVerifier → (Either ErrLog InfoLog) ⊎ Unit
+processProposal : {- NodeId → -} ProposalMsg → Epoch → ValidatorVerifier → Either (Either ErrLog InfoLog) Unit
+processProposal {- peerId -} proposal myEpoch vv =
+  case pProposal of λ where
+    (Left e) → Left (Left e)
+    (Right unit) →
```
haroldcarr, line 23: Question: is this binding `unit` or pattern matching on `unit`?

LibraBFT/Impl/Consensus/Network.agda
```diff
@@ -9,10 +9,53 @@ open import LibraBFT.ImplShared.Base.Types
...
+      grd‖ proposal ^∙ pmProposal ∙ bEpoch == myEpoch ≔
+           return unit
```
haroldcarr, line 25: change `return` to `pure`

LibraBFT/ImplShared/Consensus/Types/EpochIndep.agda
```diff
@@ -537,6 +537,8 @@ module LibraBFT.ImplShared.Consensus.Types.EpochIndep where
   vmParent : Lens VoteMsg BlockInfo
   vmParent = vmVote ∙ vVoteData ∙ vdParent
 
+  vmEpoch : Lens VoteMsg Epoch
```
haroldcarr, line 540: please add comment "getter only in Haskell"

haroldcarr (approved):
LGTM.

LibraBFT/Impl/Consensus/Network.agda
cwjnkins: > Question: is this binding `unit` or pattern matching on `unit`?

This is a pattern match. Once you use `where`, you can do pattern matching arbitrarily deeply.

haroldcarr:
FWIW:  I did see this PR while I was implementing all this exact same stuff.
Two observations:
- I failed to note I was working on them in Trello
- I only read the title of this PR - it did not say anything about Network *, so I thought I would look at it after I was done.  I think if it would have been named something like "implemented and used `Network.processProposal` and `processVote` I would have seen it.

Bottom line: my bad

LibraBFT/Impl/Consensus/Network.agda
cwjnkins: > change `return` to `pure`

Done.

LibraBFT/Impl/Consensus/Network.agda
haroldcarr: magic `where` - I'll remember that.

PR #84, opened 2021-07-26
added TODO about modeling haskell hashing; 

haroldcarr:

### DIFF
--- a/LibraBFT/ImplShared/Util/Crypto.agda
+++ b/LibraBFT/ImplShared/Util/Crypto.agda
@@ -29,21 +29,32 @@ module LibraBFT.ImplShared.Util.Crypto where
 
   open WithCryptoHash sha256 sha256-cr
 
-  -- IMPL-DIFF
-  -- 1. Cannot do equivalent of Haskell "qcParts"
-  --    because bs-concat here cannot be used because it returns BitString.
-  --    TODO-2 : Hashing in Haskell includes adding tags.
-  --             Need to model those tags and to postulate injectivity properties.
-  -- 2. No need to deal with a 'Proposal' for stable test hashes.
-  hashBD : BlockData → HashValue
-  hashBD = sha256 ∘ bs-concat ∘ blockDataBSList
-   where
-    blockDataBSList : BlockData → List ByteString
-    blockDataBSList (BlockData∙new
-                      epoch round
-                      (QuorumCert∙new voteData (LedgerInfoWithSignatures∙new ledgerInfo _sigs))
-                      blockType) =
-      encode epoch ∷ encode round ∷ encode voteData ∷ encode ledgerInfo ∷ encode blockType ∷ []
+  -- We do not yet have sufficient support to model precisely the hashing used in the Haskell code,
+  -- so it is better that we postulate its intended properties to ensure proofs have the desired
+  -- properties available.  Note that a simpler injectivity property such as
+  --
+  -- hashBD-inj' : ∀ {bd1} {bd2}
+  --             → hashBD bd1 ≡ hashBD bd2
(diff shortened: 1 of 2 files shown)


cwjnkins (approved):
LGTM

LibraBFT/ImplShared/Util/Crypto.agda
```diff
@@ -30,8 +30,11 @@ module LibraBFT.ImplShared.Util.Crypto where
   open WithCryptoHash sha256 sha256-cr
 
   -- IMPL-DIFF
-  -- 1. cannot do "qcParts" because bs-concat does returns BitString
-  -- 2. no need to deal with a 'Proposal' for stable test hashes
+  -- 1. Cannot do equivalent of Haskell "qcParts"
```
mark-moir: These changes did not quite satisfy be because the first comment still refers to things that can't be understood without seeing the Haskell code (`qcParts`) and because capitalising the first word of the second comment did not sufficiently enhance my understanding of it 😛 .

I looked at the Haskell code and tried to write comments that would satisfy me, but found deeper issues in that our Agda implementation does not accurately reflect the Haskell code, including that we don't yet include `HashTag`s when hashing.

So I have pushed a few commits that aim to explain the issues, make progress towards addressing them, and leave appropriate `TODO`s.

mark-moir: @cwjnkins please be aware of the changes I have suggested.  I suspect the lack of `HashTag`s raises some larger issues.  In particular, off the top of my head, it seems that we should also include `Block` as a `Part` when instantiating the system model, and when we do so, we will not be able to prove many of our properties without addressing the issues I've raised in the comments I've pushed, exactly because we would not be able to preclude the "masquerading" possibility mentioned in those comments.  Actually, I've pushed another commit making this point as well.

cwjnkins: > @cwjnkins please be aware of the changes I have suggested. I suspect the lack of HashTags raises some larger issues.

@mark-moir Understood, thanks for pointing this out!

LibraBFT/ImplShared/Util/Crypto.agda
```diff
@@ -29,18 +29,38 @@ module LibraBFT.ImplShared.Util.Crypto where
...
+
+  -- TODO-2: Similarly express other hashing functions and their intended injectivity properties where
+  -- the Agda implementation does not accurately reflect the Haskell implementation.
+
+  -- TODO-2: Enable support for the hashS function we use in the Haskell code, which hashes tuples
+  -- of serializable types), and then define hashBD using this and prove that it ensures hashBD-inj.
+  -- Note also the TODO below related to HashTags.
+  postulate
+    hashBD     : BlockData → HashValue
+    hashBD-inj : ∀ {bd1 bd2}
```
cwjnkins, line 52: This has a rather unwieldy return type. If you're ok with this, I would like to package all of these pieces together into a record (happy to take name suggestions!)

cwjnkins (changes requested):
@mark-moir This change request is in regards to your recent commits.

LibraBFT/ImplShared/Util/Crypto.agda
mark-moir: Good suggestion @cwjnkins, thanks.  I have offered a name and padded it out with a definition :smile:.

mark-moir:
@haroldcarr, sorry, I've rather hijacked this PR.  Feel free to merge if you're happy with it.

haroldcarr:
I am going to squash/merge.
I want to have more discussion on hashing in our next meeting (for bandwidth).

PR #88, opened 2021-07-28
"getter only" lens

haroldcarr:
This makes getter only lens "weird" again.

Note : this PR is not against main - it is against harold-RoundManager-processSyncInfoM-processLocalTimeouotM


cwjnkins (approved):
Once small typo (pushed a commit for this), otherwise looks good to me.

PR #90, opened 2021-07-28
impl SafetyRules.signTimeoutM

haroldcarr:
also: ; make RoundManager.processLocalTimeoutM use BroadcastSyncInfo

### DIFF
--- a/LibraBFT/Impl/Consensus/SafetyRules/SafetyRules.agda
+++ b/LibraBFT/Impl/Consensus/SafetyRules/SafetyRules.agda
@@ -203,8 +203,7 @@ module constructAndSignVoteM-continue2 (voteProposal : VoteProposal) (validatorS
     verifyAndUpdateLastVoteRoundM (proposedBlock ^∙ bBlockData ∙ bdRound) safetyData ∙?∙ step₁
 
   step₁ safetyData1 = do
-    lSafetyData ∙= safetyData1  -- TODO-1: resolve discussion about pssSafetyData vs lSafetyData
-    -- @CWJENKINS: replace with `extensionCheckM` (probably needs "fancier" eitherS function)
+    pssSafetyData-rm ∙= safetyData1
     pure (extensionCheck voteProposal) ∙?∙ (step₂ safetyData1)
 
   step₂ safetyData1 voteData = do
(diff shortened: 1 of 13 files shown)


LibraBFT/Impl/Consensus/SafetyRules/SafetyRules.agda
```diff
@@ -247,5 +247,33 @@ signProposalM blockData = do
...
+          bail fakeErr
+          --(ErrIncorrectPreferredRound (here []) (timeout^.toRound) (safetyData^.sdPreferredRound))
+        ‖ timeout ^∙ toRound <? safetyData ^∙ sdLastVotedRound ≔
+          bail fakeErr
+          --(ErrIncorrectLastVotedRound (here []) (timeout^.toRound) (safetyData^.sdLastVotedRound))
+        ‖ timeout ^∙ toRound >? safetyData ^∙ sdLastVotedRound ≔
+          verifyAndUpdateLastVoteRoundM (timeout ^∙ toRound) safetyData
+            ∙^∙ withErrCtx (here' [])
+            ∙?∙ (λ safetyData1 → do
+            lSafetyData ∙= safetyData1
```
mark-moir, line 266: If I've understood correctly, we define the `lSafetyData` Lens only because in the Haskell code, the (equivalent) `pssSafetyData` Lens is derived using the `RWPersistentSafetyStorage` capability, which we're not modeling in the Agda implementation.  If I've got that right, let's spell `lSafetyData` as `pssSafetyData<something>` and put a comment noting this relationship next to its definition.  `<something>` could be just `'` or `equiv` or <something else> :smile:.

mark-moir (approved):
Looks good, thanks @haroldcarr.  Just one comment on spelling/commenting.

cwjnkins (approved):
LGTM!

LibraBFT/Impl/Consensus/SafetyRules/SafetyRules.agda
cwjnkins: My preference for naming is `pssSafetyData-rm`, indicating that it is akin to the `pssSafetyData` of the Haskell code with the type for the state instantiated with `RoundManager`.

mark-moir: That works for me.

PR #94, opened 2021-07-30
Add `OutputQc∈RoundManager` to `handleProposalSpec.Contract`, use it to prove one of the `qcVoteSigsSentB4` cases

mark-moir:
Here is some progress on extending the `handleProposalSpec.Contract` to ensure `OutputQc∈RoundManager`, and then use it together with `qcVoteSigsSentB4` to prove one of the `TODO`s in `VotesOnce`.  It required generalising various properties.  I've left some postulates, `obm-dangerous-magic!`, etc., but I think they should all be provable.

### DIFF
--- a/LibraBFT/Impl/Consensus/RoundManager/Properties.agda
+++ b/LibraBFT/Impl/Consensus/RoundManager/Properties.agda
@@ -234,7 +234,7 @@ module processProposalMSpec (proposal : Block) where
                   (transNoEpochChange{i = pre}{j = st}{k = stUpdateRS} noEpochChange refl)
                   (OutputProps.++-NoProposals outs _ (OutputProps.NoMsgs⇒NoProposals outs noMsgOuts) refl)
                   (inj₂ (Voting.mkVoteSentCorrect vm recipient
-                          (OutputProps.++-NoVotes-OneVote outs _ (OutputProps.NoMsgs⇒NoVotes outs noMsgOuts) refl)
+                          (OutputProps.++-NoMsgs-OneVote outs _ noMsgOuts refl)
                           (Voting.glue-VoteGeneratedCorrect-VoteNotGenerated{s₂ = st}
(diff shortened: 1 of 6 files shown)


LibraBFT/Impl/Consensus/RoundManager/Properties.agda
```diff
@@ -294,7 +294,7 @@ module ensureRoundAndSyncUpMSpec
       -- General invariants / properties
       rmInv         : Preserves RoundManagerInv pre post
       noEpochChange : NoEpochChange pre post
-      noVoteOuts    : OutputProps.NoVotes outs
+      noVoteOuts    : OutputProps.NoMsgs outs
```
cwjnkins, line 297: Even though `ensureRoundAndSyncUpM` does not *currently* produce any messages, it very well may once syncing up is implemented (the peer might send a local message to itself indicating that there is a new epoch). I therefore think we should try to avoid using the fact that this function produces no output messages in proofs.

LibraBFT/Impl/Properties/Util.agda
```diff
@@ -69,7 +66,7 @@ module OutputProps where
     NoMsgs⇒NoVotes = proj₁ ∘ proj₂ ∘ NoMsgs⇒×
 
     OneVote : VoteMsg → List Author → Set
-    OneVote vm pids = List-filter isSendVote? outs ≡ (SendVote vm pids ∷ [])
+    OneVote vm pids = List-filter isOutputMsg? outs ≡ (SendVote vm pids ∷ [])
```
cwjnkins, line 69: @haroldcarr Is it the case that when a peer discovers that it needs to advance to a new epoch while processing a proposal, it will never generate a vote for that proposal?

cwjnkins (changes requested):
This looks good, but there is a significant issue regarding the change to the contract for `ensureRoundAndSyncUpM`.

PR #95, opened 2021-07-30
Revert changes to RoundManager output contracts introduced in PR #94

cwjnkins:
As I mentioned in PR #94, it will not be true of the final implementation that `ensureRoundAndSyncUpM` sends not output messages. I have reverted these contract changes, and many of the accompanying changes to lemmas in `LibraBFT.Impl.Properties.Util.agda` I have also repaired and simplified the case of `newVote⇒lv≡` in `LibraBFT.Impl.Properties.VotesOnce.agda` proved in PR #94 .


mark-moir:
@cwjnkins good point, good catch.  The new `sendMsg∈actions` looks useful to navigate `Action`s and `Output`s.  I won’t be looking in more detail or working on anything today, but I approve in principle, so feel free to continue on your branch and I’ll catch up on Monday.  Thanks!

mark-moir:
And now I approve in practice :smile:.  Thanks @cwjnkins.

PR #96, opened 2021-08-02
Progress on system-level invariant, `Contract`s and `Requirements` needed for `qcVoteSigsSentB4` plus progress on that

mark-moir:
Best to read one commit at a time, I think.

### DIFF
--- a/LibraBFT/Impl/IO/OBM/Properties/InputOutputHandlers.agda
+++ b/LibraBFT/Impl/IO/OBM/Properties/InputOutputHandlers.agda
@@ -39,56 +39,56 @@ module handleProposalSpec (now : Instant) (pm : ProposalMsg) where
 
   open handleProposal now pm
 
-  record Contract (pre : RoundManager) (_ : Unit) (post : RoundManager) (outs : List Output) : Set where
-    constructor mkContract
-    field
-      -- General properties / invariants
-      rmInv              : Preserves RoundManagerInv pre post
-      noEpochChange      : NoEpochChange pre post
-      -- Voting
-      voteAttemptCorrect : Voting.VoteAttemptCorrectWithEpochReq pre post outs (pm ^∙ pmProposal)
-      -- Signatures sent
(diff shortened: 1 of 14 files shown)


LibraBFT/Impl/IO/OBM/Properties/InputOutputHandlers.agda
```diff
@@ -39,56 +42,73 @@ module handleProposalSpec (now : Instant) (pm : ProposalMsg) where
...
-    contract-step₁ : _
-    proj₁ (contract-step₁ (myEpoch@._ , vv@._) refl) (inj₁ e)  pp≡Left =
-      contractBail _ refl
-    proj₁ (contract-step₁ (myEpoch@._ , vv@._) refl) (inj₂ i) pp≡Left =
-      contractBail _ refl
-    proj₂ (contract-step₁ (myEpoch@._ , vv@._) refl) unit pp≡Right =
-      processProposalMsgMSpec.contract now pm pre (Contract pre) pf
+      mSndr            : NodeId
+      m∈pool           : (mSndr , P pm) ∈ pool
+      qcs∈RmSigsSentB4 : QC.SigsForVotes∈Qc∈Rm-SentB4 pre pool
```
cwjnkins, line 49: I think this (`qcs∈RmSigsSentB4`) should be moved into `rmInv`, since the handler preserves the property between pre and post states that QC signatures in the round manager belong to the pool.

cwjnkins (approved):
This looks really good! I have one suggested change I would like to implement, that is mostly a matter of organization.

LibraBFT/Impl/IO/OBM/Properties/InputOutputHandlers.agda
mark-moir: Thanks @cwjnkins.  That was my first instinct too, but `rmInv` does not have a pool in scope: it is just about `RoundManager`s.  We could add a pool, but I think it makes sense to keep separate those properties that are purely about `RoundManager`.

PR #97, opened 2021-08-02
implement RoundState.processCertificatesM

haroldcarr:

PR #101, opened 2021-08-03
comment out logEE usage; fix comparison condition

haroldcarr:
this addresses review comments from @mark-moir for 
https://github.com/oracle/bft-consensus-agda/pull/97


mark-moir:
Thanks @haroldcarr.  I commented out the last one and fixed the proof.  Everything else looks good, so APPROVED if you're happy with my change.

PR #100, opened 2021-08-02
Progress on proving that round manager QC votes have had their signatures sent before

cwjnkins:
Updated the proof of `invariantsCorrect` now that "round manager QC votes have had their signatures sent before" is tracked as a property preserved by peer handler steps.

Also merged in `chris-handle-vote-spec` (which postulates the contract for `handleVote`, so that more progress can be made on the other cases for the `VotesOnce` property)

PR #102, opened 2021-08-03
Address review comments for #100, merge in `harold-rm-logEE-fix-condition`, add first cut of `SyncInfo.verifyM.Contract`

mark-moir:
* Addressed minor review comments for #100, including inner modules in `RoundManager` consistent with rest
* Merged in the branch for #101 (which will be merged shortly) to get `syncUpM`
* Added first cut of `SyncInfo.verifyM.Contract`

### DIFF
--- a/LibraBFT/Impl/Consensus/ConsensusTypes/SyncInfo/Properties.agda
+++ b/LibraBFT/Impl/Consensus/ConsensusTypes/SyncInfo/Properties.agda
@@ -1,0 +1,53 @@
+{- Byzantine Fault Tolerant Consensus Verification in Agda, version 0.9.
+
+   Copyright (c) 2021, Oracle and/or its affiliates.
+   Licensed under the Universal Permissive License v 1.0 as shown at https://opensource.oracle.com/licenses/upl
+-}
+
+open import LibraBFT.Base.Types
+open import LibraBFT.Concrete.System
+open import LibraBFT.Concrete.System.Parameters
+open import LibraBFT.Impl.Properties.Util
+open import LibraBFT.Impl.Types.BlockInfo as BI
+open import LibraBFT.ImplShared.Consensus.Types
+open import LibraBFT.ImplShared.Consensus.Types.EpochDep
+open import LibraBFT.ImplShared.Interface.Output
+open import LibraBFT.ImplShared.Util.Util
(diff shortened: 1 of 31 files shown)


LibraBFT/Impl/Consensus/ConsensusTypes/SyncInfo/Properties.agda
```diff
@@ -0,0 +1,53 @@
...
+open import Optics.All
+
+open RoundManagerInvariants
+open RoundManagerTransProps
+
+module LibraBFT.Impl.Consensus.ConsensusTypes.SyncInfo.Properties where
+
+module verifyMSpec (self : SyncInfo) (validator : ValidatorVerifier) where
+
+  record Requirements (pre : RoundManager) : Set where
```
cwjnkins, line 27: This would be a very awkward requirement, since it would have to be bubbled up to all callers of `verifyMSpec` -- and we handle this already with `RoundManagerInv`.

If there are some properties of `SIVerifyProps` that require this in order to be proved, I think it's better to make the final guarantee be a conditional (see below)

LibraBFT/Impl/Consensus/ConsensusTypes/SyncInfo/Properties.agda
```diff
@@ -0,0 +1,53 @@
...
+  module _ (pre : RoundManager) (reqs : Requirements pre) where
+
+   record Contract (r : Either ErrLog Unit) (post : RoundManager) (outs : List Output) : Set where
+     constructor mkContract
+     field
+       -- General properties / invariants
+       noStateChange : pre ≡ post
+       noMsgOuts     : OutputProps.NoMsgs outs
+       -- Syncing
+       syncResCorr   : r ≡ Right unit → SIVerifyProps pre post (rmc reqs)
```
cwjnkins, line 53: Here, `syncResCorr` can take an additional argument `reqs : Requirements pre` (or just `rmc : RoundManager-correct pre`), instead of making it a requirement of the entire contract. As a bonus, that way we do not need to know that the round manager is correct just to get that the pre and post state have not changed.

cwjnkins (approved):
I will move forward with the change to the specification for `verifyM` to illustrate what I mean, then we can discuss later what we think is the best style for stating contracts.

PR #103, opened 2021-08-03
implement more verify/validate

haroldcarr:
various verify/validate/wellFormed, such as:

Block.verifyWellFormed
Block.validateSignature

Vote.verify
Vote.addTimeoutSignature

ProposalMsg.verifyWellFormed
ProposalMsg.verify

TimeoutCertificate.verify

SyncInfo.verify

Note: this postulates : payloadIsEmpty : when we do Epoch change, we will need to decide what empty is.

PR #106, opened 2021-08-04
use maybe-RWST

haroldcarr:
PR 
https://github.com/oracle/bft-consensus-agda/pull/103 
renamed `maybeS` to `maybeS-RWST`
and defined a new general "monadic" `maybeS`

This PR replaces some existing uses of `maybeS` in the `LBFT` monad with `maybeS-RWST`

IMPORTANT: See the comment in `Syntax.agda` on when to use the `-RWST` variant.

### DIFF
--- a/LibraBFT/Impl/Consensus/SafetyRules/SafetyRules.agda
+++ b/LibraBFT/Impl/Consensus/SafetyRules/SafetyRules.agda
@@ -96,7 +96,7 @@ verifyAndUpdatePreferredRoundM quorumCert safetyData = do
 verifyAuthorM : Maybe Author → LBFT (Either ErrLog Unit)
 verifyAuthorM author = do
   vs ← use (lSafetyRules ∙ srValidatorSigner)
-  maybeS-RWST vs (bail fakeErr) {-(ErrL (here' ["srValidatorSigner", "Nothing"]))-} $ λ validatorSigner →
+  maybeS vs (bail fakeErr) {-(ErrL (here' ["srValidatorSigner", "Nothing"]))-} $ λ validatorSigner →
     maybeS-RWST
       author
       (bail fakeErr) -- (ErrL (here' ["InvalidProposal", "No author found in the proposal"])))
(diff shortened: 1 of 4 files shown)


mark-moir:
Looks good @haroldcarr.  I suggest adding comments to the two places you didn't change explaining that we could use `maybeS-RWST` but we don't because we have already proved properties about these functions expressed using `maybeS`.  It would be instructive to change them and fix the proofs, but I agree it's not a priority now.  APPROVED.

LibraBFT/Impl/Consensus/SafetyRules/SafetyRules.agda
```diff
@@ -96,8 +96,8 @@ verifyAndUpdatePreferredRoundM quorumCert safetyData = do
 verifyAuthorM : Maybe Author → LBFT (Either ErrLog Unit)
 verifyAuthorM author = do
   vs ← use (lSafetyRules ∙ srValidatorSigner)
-  maybeS vs (bail fakeErr) {-(ErrL (here' ["srValidatorSigner", "Nothing"]))-} $ λ validatorSigner →
-    maybeS
+  maybeS-RWST vs (bail fakeErr) {-(ErrL (here' ["srValidatorSigner", "Nothing"]))-} $ λ validatorSigner →
```
cwjnkins, line 99: It isn't necessary to use `maybeS-RWST` here, and even slightly preferable to use `maybeS`. This is generally true whenever the "scrutinee" (whatever we are doing case analysis on -- here `vs`) is exactly one of the variables that have been bound by a previous sequence.

LibraBFT/Impl/Consensus/SafetyRules/SafetyRules.agda
```diff
@@ -96,8 +96,8 @@ verifyAndUpdatePreferredRoundM quorumCert safetyData = do
...
+    maybeS-RWST
```
cwjnkins, line 100: For comparison, this is an example where `maybeS-RWST` _is_ preferred, since the scrutinee is `author` (an argument to the function).

cwjnkins (changes requested):

LibraBFT/Impl/Consensus/SafetyRules/SafetyRules.agda
mark-moir: Thanks for this comment, the examples, and #108.  It would be good to capture the considerations here in comments more generally.  Is the following accurate/adequate?

> Within the `RWST` monad, `if` and `if-RWST` are semantically interchangeable, and similarly for `maybeS` and `maybe-RWST`.  The difference is in how proof obligations are generated, with the `*-RWST` variants generating new `weakestPre` obligations for each case.  In some cases, this is helpful for structuring proofs, while in other cases it is unnecessary and introduces more structure to the proof without adding any benefit.  A rule of thumb is that, if the ""scrutinee" (whatever we are doing case analysis on, i.e., the first argument) is the value provided via `RWST-bind` by a previous code block, then we already have a `weakestPre` proof obligation, so introducing additional ones via the `*-RWST` variants only creates more work and provides no additional benefit.

cwjnkins: Yes, that's an excellent description.

mark-moir: Thanks for confirming @cwjnkins.  @haroldcarr, could you include the comment above before merging?  I'm thinking maybe in `RWST.Syntax`?

haroldcarr: I have added that description to Syntax.agda.

PR #108, opened 2021-08-04
LibraBFT: Impl: Consensus: SafetyRules: SafetyRules: use `maybeS` when scrutinee is variable bound by a `>>=`

cwjnkins:


haroldcarr (approved):
YES

PR #109, opened 2021-08-05
Reworking contracts for proving qcVoteSigsSentB4

cwjnkins:
I did not finish this, but just FYI: I think `invariantsCorrect` in `LibraBFT/Impl/Handle/Properties.agda` needs to be restated so that the votes in QCs in a peer round manager have come from the pool of the _preceding_ system state (I have updated the peer contracts to guarantee that QCs in output messages are in the post state).

There's also a useful lemma in here for relating `OutputProps.NoMsgs` to `QCProps.OutputQc∈RoundManager`

### DIFF
--- a/LibraBFT/Impl/Handle/Properties.agda
+++ b/LibraBFT/Impl/Handle/Properties.agda
@@ -51,6 +51,7 @@ initRMSatisfiesInv =
     (mkSafetyRulesInv (mkSafetyDataInv refl z≤n))
 
 invariantsCorrect -- TODO-1: Decide whether this and direct corollaries should live in an `Properties.Invariants` module
+                  -- TODO-2: This needs to be tightend so that the msgPool is of the previous system state (if there is one)
   : ∀ pid (pre : SystemState)
     → ReachableSystemState pre → RoundManagerInv (msgPool pre) (peerStates pre pid)
 invariantsCorrect pid pre@._ step-0 = initRMSatisfiesInv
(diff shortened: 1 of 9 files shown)


LibraBFT/Impl/Handle/Properties.agda
```diff
@@ -51,6 +51,7 @@ initRMSatisfiesInv =
     (mkSafetyRulesInv (mkSafetyDataInv refl z≤n))
 
 invariantsCorrect -- TODO-1: Decide whether this and direct corollaries should live in an `Properties.Invariants` module
+                  -- TODO-2: This needs to be tightend so that the msgPool is of the previous system state (if there is one)
```
mark-moir, line 54: I don't think this is right or makes sense.  This is an invariant (as opposed to a transition property), so what previous state?  I think we will be *applying* the invariant to the *prestate* of some transitions, such as in `VotesOnce`, which makes sense to me.  Let's discuss at higher bandwidth.

mark-moir (approved):
@cwjnkins, thanks.  I haven't looked in great detail, and am out of time for today.  I've just made one comment we can discuss.  There are no conflicts, so I'm happy for you to merge either this or #110 as you see fit.  APPROVED.

LibraBFT/Impl/Handle/Properties.agda
cwjnkins: >  I think we will be applying the invariant to the prestate of some transitions, such as in `VotesOnce`, which makes sense to me. Let's discuss at higher bandwidth.

I agree that we should discuss at a higher bandwidth. The motivation for these changes is that I'm not sure that it makes sense to say that QCs in outputs came from the prestate and not the poststate. I will finish the changes I have in mind later today so we have something concrete to consider.

PR #111, opened 2021-08-05
Reworked `qcVoteSigsSentB4` proof, progress on contract for `ensureRoundAndSyncUpM`

cwjnkins:
This PR clarifies what I meant by my earlier comment that the message pool mentioned in the type of `RoundManagerInv`, in the proof of `invariantsCorrect`, should be the pool _prior_ to the last step of the system. It also contains progress and some cleanup for the parts of the peer handler contracts used to prove `qcVoteSigsSentB4`

### DIFF
--- a/LibraBFT/Impl/Consensus/RoundManager/Properties.agda
+++ b/LibraBFT/Impl/Consensus/RoundManager/Properties.agda
@@ -309,12 +309,7 @@ module ensureRoundAndSyncUpMSpec
 
   open ensureRoundAndSyncUpM now messageRound syncInfo author helpRemote
 
-  -- TODO-2: Add requirements capturing that any signatures in any QCs in the SyncInfo have been
-  -- sent before.  Probably want to refactor properties in LibraBFT.Impl.Properties.Util.QC
-  postulate
-    Requirements : SentMessages → SyncInfo → Set
-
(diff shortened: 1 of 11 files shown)


LibraBFT/Impl/Consensus/RoundManager/Properties.agda
```diff
@@ -309,47 +309,50 @@ module ensureRoundAndSyncUpMSpec
...
       field
         -- General invariants / properties
-        rmInv         : Preserves (RoundManagerInv pool) pre post
+        rmInv         : Preserves RoundManagerInv pre post
         noEpochChange : NoEpochChange pre post
         noVoteOuts    : OutputProps.NoVotes outs
         -- Voting
         noVote        : VoteNotGenerated pre post true
+        -- Signatures
+        outQcs∈RM     : QCProps.OutputQc∈RmOrMsg outs post {!!}
```
mark-moir, line 324: I think this should be a similar property that requires either the `QC` is in the `RoundManager` or  it is in `SyncInfo`.  We know at call sites that the `SyncInfo` has come from a message that has been sent.

mark-moir (approved):
Thanks @cwjnkins.  There's a lot going on here.  I have one comment, which I will explore after merging and will report back.

PR #113, opened 2021-08-06
Sketch different approach to tracking provenance of signatures in output QCs

mark-moir:
Sorry this is rushed.

It sketches a different way to track provenance of signatures in output QCs, so that each function's doesn't have to know about things outside it (such as network messages, message pools, etc.)

The key idea is to replace `OutputQc∈RmOrMsg` by `OutputQc∈RmOr`, which takes a predicate that must hold for a `QC` that is in the poststate but not in the prestate.  By making this a predicate over pre and post states, it becomes simple to stitch these together up and down the call stack, so we can relate `QC`s to what we know about (with respect to the whole step's prestate and the message pool) them near the top level.

Very much a sketch, but hope the idea is clear/useful.

### DIFF
--- a/LibraBFT/Impl/Properties/Util.agda
+++ b/LibraBFT/Impl/Properties/Util.agda
@@ -104,11 +104,14 @@ module QCProps where
       syncInfo∈msg : syncInfo SyncInfo∈NM msg
     open MsgRequirements msgReqs
 
-  data _∈RoundManager_ (qc : QuorumCert) (rm : RoundManager) : Set where
-    inHQC : qc ≡ rm ^∙ lBlockStore ∙ bsInner ∙ btHighestQuorumCert → qc ∈RoundManager rm
-    inHCC : qc ≡ rm ^∙ lBlockStore ∙ bsInner ∙ btHighestCommitCert → qc ∈RoundManager rm
+  data _∈BlockTree_ (qc : QuorumCert) (bt : BlockTree) : Set where
+    inHQC : qc ≡ bt ^∙ btHighestQuorumCert → qc ∈BlockTree bt
+    inHCC : qc ≡ bt ^∙ btHighestCommitCert → qc ∈BlockTree bt
     -- NOTE: When `need/fetch` is implemented, we will need an additional
     -- constructor for sent qcs taken from the blockstore.
+
+  _∈RoundManager_ : (qc : QuorumCert) (rm : RoundManager) → Set
+  qc ∈RoundManager rm =  qc ∈BlockTree (rm ^∙ lBlockStore ∙ bsInner)
 
   OutputQc∈RmOrMsg : List Output → RoundManager → NetworkMsg → Set
   OutputQc∈RmOrMsg outs rm msg =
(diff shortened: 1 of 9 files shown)


cwjnkins (approved):
This looks great, thanks!

LibraBFT/Impl/Properties/Util.agda
```diff
@@ -104,16 +104,24 @@ module QCProps where
...
+  qc ∈RoundManager rm =  qc ∈BlockTree (rm ^∙ lBlockStore ∙ bsInner)
+
   OutputQc∈RmOrMsg : List Output → RoundManager → NetworkMsg → Set
   OutputQc∈RmOrMsg outs rm msg =
     All (λ out → ∀ qc nm → qc QC∈NM nm → nm Msg∈Out out → qc ∈RoundManager rm ⊎ qc QC∈NM msg) outs
 
+  OutputQc∈RmOr : List Output → RoundManager → RoundManager
+                → (Q : QuorumCert → Set) → Set
+  OutputQc∈RmOr outs pre post Q =
+    All (λ out → ∀ qc nm → qc QC∈NM nm → nm Msg∈Out out → qc ∈RoundManager post → qc ∈RoundManager pre ⊎ Q qc) outs
```
cwjnkins, line 123: When attempting to repair the proof for `VotesOnce`, I encountered a problem with this formulation of the property. The contracts do not guarantee that `qc`s that appear in output messages are in the post state, so as stated this property cannot be used to obtain the disjunction that the `qc`s appear in the round manager or `Q qc`.

cwjnkins (changes requested):

LibraBFT/Impl/Properties/Util.agda
mark-moir: I agree, the sketch only included what we need to prove that QCs in  the post state were either already in the prestate or are in the message being processed.  We still need to propagate a property like you already have in some of the higher level properties: that QCs in the outputs are in the poststate (we originally said prestate, but you pointed out this does not hold), and also that signatures represented in the prestate QCs have been sent before.  That makes me think that we should have kept `RoundManagerInv` dependent on a message pool.  What do you think?

mark-moir: One more thing I forgot to say yesterday is that I think we will sometimes want to know that `RiundManagerInv` holds in the prestate in order to prove other parts of the contract.  An example in this context is using the invariant referred to above, that signatures represented in the prestate have been sent before to prove that signatures in outputs have been sent before.  `Preserves (RoundManagerInv pool)` provides that only for proving that `RoundManagerInv` holds in the poststate, not the other `Contract` properties.  This makes me think we should refactor the `Contracts`s so that we get to use `RoundManagerInv pool pre` when proving all of the fields.

cwjnkins: > We still need to propagate a property like you already have in some of the higher level properties: that QCs in the outputs are in the poststate (we originally said prestate, but you pointed out this does not hold)

This is what I'm currently working on, and it has resulted in a significant simplification: contracts only need to show that qcs in outputs have come from the post state, and that the handler preserves the property of signatures for votes being sent before between the pre and post state (assuming the received message is the pool).

> This makes me think we should refactor the Contractss so that we get to use RoundManagerInv pool pre when proving all of the fields.

Not all of the fields need this though. Where it is needed, it could be made an argument to the field that depends upon this property (and if multiple properties depend on it, they can be bundled into a record that is made the return type of a field which takes an assumption that the invariant holds for the prestate).

mark-moir:
The discussion was useful and we have continued with its outcome in other branches (`chris2-mark-qcsb4` and `mark2-chris-qcsb4`) so I am closing this one.

PR #118, opened 2021-08-09
Progress on contract for `executeAndInsertBlockE`

cwjnkins:
Also some refinements to the contracts for `RoundManager` peer handlers, and a small cleanup in one of the `Handler` proofs.


mark-moir (approved):
Good progress @cwjnkins, thanks.  I just made one observation, but I'm merging as-is.

PR #120, opened 2021-08-10
StateComputer.compute state/exceptions

haroldcarr:
added comment about StateComputer.compute not throwing exception nor maintaining state (but the possibility these may happen in the future)

Changed existing exception `ErrBlockNotFound` to `ErrCBlockNotFound` - that is used when the consensus level cannot find an expected block.

Added `ErrECCBlockNotFound` - that would be used inside the state computer, inside its "execution correction client".  It is never thrown in the Agda code (nor in the Haskell code).   But it is caught in `BlockStore.executeAndInsertBlockM`.  See the comment there.

### DIFF
--- a/LibraBFT/Impl/Consensus/BlockStorage/BlockStore.agda
+++ b/LibraBFT/Impl/Consensus/BlockStorage/BlockStore.agda
@@ -63,7 +63,7 @@ commitM finalityProof = do
     let blockIdToCommit = finalityProof ^∙ liwsLedgerInfo ∙ liConsensusBlockId
     case getBlock blockIdToCommit bs of λ where
       nothing →
-        bail (ErrBlockNotFound blockIdToCommit)
+        bail (ErrCBlockNotFound blockIdToCommit)
       (just blockToCommit) →
         ifM‖ blockToCommit ^∙ ebRound ≤?ℕ bsr ^∙ ebRound ≔
              bail fakeErr -- "commit block round lower than root"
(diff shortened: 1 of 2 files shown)


LibraBFT/Impl/Consensus/BlockStorage/BlockStore.agda
```diff
@@ -142,8 +142,13 @@ executeAndInsertBlockE bs0 block =
...
       eb ← case executeBlockE bs0 block of λ where
         (Right res) → Right res
-        (Left (ErrBlockNotFound parentBlockId)) → do
+        -- OBM-LBFT-DIFF : This is never thrown in OBM.
+        -- Because it is thrown by StateComputer in Rust (but not in OBM).
+        (Left (ErrECCBlockNotFound parentBlockId)) → do
           eitherS (pathFromRoot parentBlockId bs0) Left $ λ blocksToReexecute →
+            -- OBM-LBFT-DIFF : OBM StateComputer does NOT have state.
+            -- If it ever does have state then the following 'forM' will
+            -- need to change to some sort of 'fold'.
```
mark-moir, line 151: I would add something like: "because `executeBlockE` would change the state, so the state passed to `executeBlockE` in the `Right` case below" would no longer be `bs0`.

mark-moir (commented):
I made one metacomment, but I am confused about the other changes that are not mentioned/explained in the PR description.  Probably better at higher bandwidth.

LibraBFT/Impl/Consensus/BlockStorage/BlockStore.agda
haroldcarr: DONE.

haroldcarr: I updated the PR description to describe the other changes.

PR #119, opened 2021-08-10
Progress on tracking QCs in contracts (`executeAndInsertBlockM`)

cwjnkins:
- merge in `main`
- refine and partially prove contract for `executeAndInsertBlockM`
- new relation `_≈Block_` for block equivalence modulo signatures
- refinements to `RoundManager` contracts to track QCs

PR #121, opened 2021-08-11
Progress on `insertQuorumCertE Spec`, generalise `∈Post⇒∈PreOr`, eliminate unnecessary pools

mark-moir:
Significant progress on `insertQuorumCertE Spec` but I've left one proof that's similar to some others, suggesting a general lemma may be worthwhile, and one case where I'm stuck and could use some input and/or sleep.  I've left some notes about where I'm up to.

Also generalised `∈Post⇒∈PreOr` so we can use similar functionality further down the stack.  This required reordering some parameters, which has caused some conflicts with #119.

Also eliminated `pool` module parameters where they are unnecessary.

### DIFF
--- a/LibraBFT/Impl/Consensus/BlockStorage/Properties/BlockStore.agda
+++ b/LibraBFT/Impl/Consensus/BlockStorage/Properties/BlockStore.agda
@@ -194,7 +194,7 @@ module executeAndInsertBlockMSpec (b : Block) where
 module insertSingleQuorumCertMSpec
   (qc : QuorumCert) where
 
-  module _ (pool : SentMessages) (pre : RoundManager) where
+  module _ (pre : RoundManager) where
 
     record Contract (r : Either ErrLog Unit) (post : RoundManager) (outs : List Output) : Set where
       constructor mkContract
@@ -206,8 +206,8 @@ module insertSingleQuorumCertMSpec
         -- Voting
         noVote        : VoteNotGenerated pre post true
         -- Signatures
-        qcSigsB4 : QCProps.QCRequirements pool qc
-                   → Preserves (QCProps.SigsForVotes∈Rm-SentB4 pool) pre post
+        qcSigsB4      : ∀ {pool} → QCProps.QCRequirements pool qc
+                        → Preserves (QCProps.SigsForVotes∈Rm-SentB4 pool) pre post
 
 
     postulate -- TODO-2: prove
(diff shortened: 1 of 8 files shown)


LibraBFT/Impl/Consensus/BlockStorage/Properties/BlockStore.agda
```diff
@@ -207,8 +206,8 @@ module insertSingleQuorumCertMSpec
         -- Voting
         noVote        : VoteNotGenerated pre post true
         -- Signatures
-        qcSigsB4 : QCProps.QCRequirements pool qc
-                   → Preserves (QCProps.SigsForVotes∈Rm-SentB4 pool) pre post
+        qcSigsB4      : ∀ {pool} → QCProps.QCRequirements pool qc
+                        → Preserves (QCProps.SigsForVotes∈Rm-SentB4 pool) pre post
```
cwjnkins, line 210: Just a comment - we intend to remove the dependence on `pool` here too, right?

LibraBFT/Impl/Consensus/BlockStorage/Properties/BlockTree.agda
```diff
@@ -50,15 +53,108 @@ module insertBlockESpec (block : ExecutedBlock) (bt : BlockTree) where
...
+    pf2 : Ok' bt1 il (step₂ blockId block) → ∃[ hcb ](just hcb ≡ bt0 ^∙ btHighestCertifiedBlock × Ok' bt1 il (step₃ blockId block hcb))
+    pf2 isOk
+       with bt0 ^∙ btHighestCertifiedBlock
+    ...| just hcb   = hcb , refl , isOk
+  ...| hcb , lookupJust₂ , step3-ok
+    with          bt0 & btHighestCertifiedBlockId ∙~ block ^∙ ebId &    btHighestQuorumCert ∙~ qc
+       | inspect (bt0 & btHighestCertifiedBlockId ∙~ block ^∙ ebId &_) (btHighestQuorumCert ∙~ qc)
+       | fakeInfo ∷ []
+  ...| bt₃-true | [ refl ] | il₃-true
+    with  if ⌊ (block ^∙ ebRound) >? (hcb ^∙ ebRound) ⌋ then bt₃-true else   bt0               | inspect
```
cwjnkins, line 97: I think you should try simplifying to just doing analysis on

`(block ^∙ ebRound) >? (hcb ^∙ ebRound)`

and other such expressions. You will then get the equal proof (not just a boolean), which may help you avoid using `inspect`.  The proof state will then move forward once you inspect the result of the decision (i.e., it will be able to take the `then` or `else` branch depending on the result)

cwjnkins (approved):
LGTM, just one question and some feedback.

PR #126, opened 2021-08-12
Refine and prove `pf3`, comment on issue with defining `il3`

mark-moir:
@cwjnkins. I made progress by refining and proving `pf3`, but I did not prove its requirements.  These can presumably be proved, but I couldn't quickly/easily do it.  I think the way I defined ( `bt₃` and) `il₃` has caused a lot of unnecessary pain, so I am stopping here.  I think this provides good motivation for the weakest precondition infrastructure you have suggested for the `Error` monad :smile:.  Feedback/input welcome.


cwjnkins:
I agree -- it definitely provides good motivation for changing how we handle proofs running in the error monad. See PR #128 for an example of how things might change.

PR #128, opened 2021-08-12
Weakest precondition infrastructure for `Error`

cwjnkins:
- example proof in `LibraBFT/Impl/Consensus/BlockStorage/Properties/BlockStore.agda`
- `LibraBFT/ImplShared/Util/` has been re-arranged so that common definitions for conditionals can be shared between `RWST` and `Error`

There are still some things that could be cleaned up, in particular the current need for `fromEither` -- we can discuss how to deal with this at a higher bandwidth.

### DIFF
--- a/LibraBFT/ImplShared/Util/Dijkstra/Error/Syntax.agda
+++ b/LibraBFT/ImplShared/Util/Dijkstra/Error/Syntax.agda
@@ -17,18 +17,30 @@ private
 
 -- From this instance declaration, we get _<$>_, pure, and _<*>_ also.
 instance
-  Monad-Error : ∀ {E : Set} → Monad (Error E)
-  Monad.return Monad-Error = Error-return
-  Monad._>>=_  Monad-Error = Error-bind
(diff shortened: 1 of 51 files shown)


LibraBFT/ImplShared/Util/Dijkstra/Error/Syntax.agda
```diff
@@ -0,0 +1,46 @@
+{- Byzantine Fault Tolerant Consensus Verification in Agda, version 0.9.
```
haroldcarr, line 1: To clarify - the outer module should be `EitherD` and the `Error.agda` file should be `EitherD.agda`

haroldcarr:
Nice!

LibraBFT/ImplShared/Util/Dijkstra/Error/Syntax.agda
cwjnkins: Done!

mark-moir:
Closing because it's obsolete (the work here is now merged in more recent branches).  Great work though :smile:.

PR #131, opened 2021-08-17
QC vote signatures sent before, weakest precondition infrastructure for error-throwing peer handlers

cwjnkins:
This PR contains (still in-progress) work on the following two big items

- Progress on the `VotesOnce` proof:
  - Completed proof of major lemma, `qcVoteSigsSentB4`, stating signatures for votes represented in quorum certificates in a peer's round manager have been sent before
  - Progress proving properties of peer handlers needed to show `qcVoteSigsSentB4` 

- Weakest precondition machinery for `EitherD`, a type constructor for peer handler code that possibly throws errors (but does not use the full `RWS` functionality)

It also contains documentation (see `docs/PeerHandlerContracts.org`) on getting started proving peer handler contracts.

### DIFF
--- a/LibraBFT/Base/KVMap.agda
+++ b/LibraBFT/Base/KVMap.agda
@@ -57,6 +57,7 @@ module LibraBFT.Base.KVMap  where
                   → lookup k kvm ≡ nothing
                   → KVMap Key Val
    elems          : KVMap Key Val → List Val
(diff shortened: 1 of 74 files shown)


mark-moir:
Fantastic work, Chris!  Thanks for everything and please stay in touch!!

PR #133, opened 2021-08-19
Experiment with streamlining boilerplate for `EitherD` variants

mark-moir:
Experimental approach to streamlining boilerplate for `EitherD` variants, as recommended by @cwjnkins [here](https://github.com/oracle/bft-consensus-agda/blob/main/docs/PeerHandlerContracts.org#eitherd-and-monadic-bind).

Pros:
* Avoids repeating type signatures several times to get all variants

Cons:
* ~~requires adding `.E` to use ordinary variant (the first commit avoids this, but is a bit less streamlined)~~  UPDATE: we can have functions and modules with the same name, and most/many functions will be called only/predominantly in one form, so we can add a "default" for each function, and don't even need to include the streamlined type signature.

To discuss/explore:
* How well does this work for existing code and proofs?  A good test case is `insertBlock*`.  ~~I have not tried.~~  UPDATE: I have applied this to more cases and it works well, did not break any proofs, shows flexibilty to use different variants in different contexts.  I added some comments and updated `PeerHandlerContracts.org`.


cwjnkins (approved):
LGTM!

PR #147, opened 2021-08-31
Start refining `insertBlockESpec`

mark-moir:
This is the first of two closely related PRs.  This one refines the contract for `insertBlockESpec`, providing it with `Requirements` that will be needed to prove its contract.  The [next one](#148) refactors that, so that each part of the contract is provided with only the prerequisite properties that its proof is expected to require.

This one also:
* reinstates the `stateComputerResult` that broke the proof in`executeAndInsertBlockE`, fixes that proof, and documents it in a little more detail for "tutorial" purposes; also updates the description in `PeerHandlerContracts.org` accordingly.
* Establishes some utility types for more clearly expressing requirements relating `Block`s and their ids.
* Refines the notion of two `Block`s being "equivalent" (`_≈Block_`), so that it does not require them to have the same `QC` signatures
* Adds initial machinery for ` EitherD-⇒` analogous to ` RWS-⇒` (not used yet)
* Various other small changes and fixes

### DIFF
--- a/LibraBFT/Impl/Consensus/BlockStorage/Properties/BlockStore.agda
+++ b/LibraBFT/Impl/Consensus/BlockStorage/Properties/BlockStore.agda
@@ -31,18 +31,6 @@ open QCProps
 
 module LibraBFT.Impl.Consensus.BlockStorage.Properties.BlockStore where
 
-module getBlockSpec (hv : HashValue) (bs : BlockStore) where
-
-  Ok : Set
-  Ok = ∃[ eb ] (getBlock hv bs ≡ just eb)
-
-  postulate -- TODO-2: Refine contract
-            -- This contract is only a sketch, and will need to be modified when
-            -- hash collisions are are modeled.
-    correctBlockData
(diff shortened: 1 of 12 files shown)


LibraBFT/Impl/Consensus/BlockStorage/Properties/BlockStore.agda
```diff
@@ -180,7 +172,7 @@ module executeAndInsertBlockMSpec (b : Block) where
         → LBFT-weakestPre (executeAndInsertBlockM b) Post pre
     proj₁ (contract Post pfBail pfOk ._ refl) e ≡left = pfBail e
     proj₂ (contract Post pfBail pfOk ._ refl) (bs' , eb) ≡right ._ refl unit refl
-       with executeAndInsertBlockESpec.contract bs b
+       with executeAndInsertBlockESpec.contract bs b (obm-dangerous-magic' "propagate requirements up")
```
haroldcarr, line 175: hmm

haroldcarr (approved):
Cool that StateComputer is back!

mark-moir:
I'm reopening this, and will then create another one on top of it.  Together, these replace #153.

PR #148, opened 2021-08-31
Refactor `insertBlockESpec` requirements and start propagating them

mark-moir:
This builds on #147, refactoring the `Requirement`s provided to the proof for `insertBlockESpec.contract*` so that each component of the `Contract` receives only the `Requirement` its proof is expected to need.  These requirements are then propagated up.

There are requirements relating a `Block` to be inserted to its ID (specifically, a `Block`s ID should be the hash of its `BlockData`), and these have been propagated all the way up to `InputOutputHandler`, where the required check is made.  

Other requirements concern similar properties for `Block`s stored in the `BlockTree` and the beginning of propagating assumptions about hash collisions towards a context in which they can be tied to a specific reachable system state.  The latter two requirements have not yet been propagated all the way up to where they can be provided, and are postulated (via `obm-dangerous-magic'` for now).

This experiment makes me think this is the right way to go, but I am keeping two separate pull requests for comparison and discussion.  I propose to merge and squash this one, and forget the initial steps represented by #147.

### DIFF
--- a/LibraBFT/Impl/Consensus/BlockStorage/Properties/BlockStore.agda
+++ b/LibraBFT/Impl/Consensus/BlockStorage/Properties/BlockStore.agda
@@ -63,12 +63,14 @@ module executeAndInsertBlockESpec (bs0 : BlockStore) (block : Block) where
   Err = ∃[ e ] (executeAndInsertBlockE bs0 block ≡ Left e)
   ------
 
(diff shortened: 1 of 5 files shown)


haroldcarr (approved):
LGTM!

(actually, looks GREEK to me!)

PR #149, opened 2021-08-31
types

haroldcarr:
The main thing this PR request does is move most of the types from `EpochManagerTypes` into `EpochIndep`.

It also rearranges the existing types in `EpochIndep` to be in the same order as Haskell.

I suggest looking at the changes in commit order - because the last one just cleans up import statements and touches many files.


mark-moir:
LGTA (Looks Good To Agda :smile:).  APPROVED.  Thanks @haroldcarr.

PR #153, opened 2021-09-02
Feed requirements to `insertBlockESpec` from network layer, reinstate `stateComputeResult`

mark-moir:
This replaces #147 and #148, connecting information validated at the network layer all the way down to the leaves of the call tree where it is needed.  This is now achieved via dependent module parameters, rather than explicitly in `Contract`s as in previous experiments; this is more readable and useable.  More specifically, we need to relate a `Block` to be inserted to its ID (specifically, a `Block`s ID should be the hash of its `BlockData`).

Other requirements concern similar properties for `Block`s stored in the `BlockTree` and the beginning of propagating assumptions about hash collisions towards a context in which they can be tied to a specific reachable system state.  A start has been made by feeding `RoundManagerInv` down the stack but this is not yet connected to assumptions about hash collisions, and is postulated (via `obm-dangerous-magic'`) for now.

This PR also:
* reinstates `stateComputeResult` and fixes and documents the proof that it previously broke.
* refines `≈Block` to also ignore differences in signatures included in `QC`s, not just the `Block` signature
* organises LibraBFT.Impl.Properties.Util` somewhat better

### DIFF
--- a/LibraBFT/Base/KVMap.agda
+++ b/LibraBFT/Base/KVMap.agda
@@ -89,6 +89,11 @@ module LibraBFT.Base.KVMap  where
                   → (prf : lookup k kvm ≡ nothing)
                   → lookup k (kvm-insert k v kvm prf) ≡ just v
 
+   -- Haskell's insert updates the value even if the key is already in the map
+   lookup-correct-haskell
+                  : {kvm : KVMap Key Val}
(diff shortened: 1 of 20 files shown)


mark-moir:
I'm closing this and will shortly create another one to replace it.

PR #158, opened 2021-09-09
Start towards properties depending on *specific* hash collisions found by peers vs. mere *existence* of hash collisions

mark-moir:
So far, our proofs depend on an assumption that hash collisions don't exist.  But they do!  We have required hash collisions to be put in context of a `ReachableSystemState` just to remind ourselves that we cannot assume lack of hash collisions "for free".  But this did not mean much because nothing connected the specific hash collisions to that `ReachableSystemState`.

This PR makes the first steps towards doing so, dealing with one important class of potential hash collisions: those on `BlockId`s, requiring properties to depend only on the assumption that these hash collisions do not exist in the specific state for which a property is proved.  This is achieved by defining a notion of a hash being in the state (e.g., a `Proposal` that has the same hash as a `Block` already stored in the `BlockTree`, but with different `BlockData`.  See `LibraBFT.ImplShared.Util.HashCollisions.HashCollisionFound`.

This requires us to capture the fact that we validate new `Proposal`s "off the wire" before handling them, and propagate this information "down the stack" in order to connect it with an assumption that a hash collision does not exist *in the `ReachableSystemState` in which the `Proposal` is handled.  This validity of `Proposal`s is propagated down via module parameters, and the lack of hash collisions is propagated through arguments to the specific contract properties that require them.

There is room for improvement streamlining this, especially in `LibraBFT.Impl.Properties.VotesOnce`.  There are also still a number of uses of `meta-sha256-cr`, which does not tie the hash collision to the `ReachableSystemState`.  Addressing these will likely require the addition of more kinds of specific hash collision.  When we have eliminated all uses of `meta-sha256-cr`, and proved all properties, we will then know that we have a precise and narrow summary of the possible hash collisions, the absence of which our proofs require.

### DIFF
--- a/LibraBFT/Impl/Consensus/BlockStorage/Properties/BlockStore.agda
+++ b/LibraBFT/Impl/Consensus/BlockStorage/Properties/BlockStore.agda
@@ -44,7 +44,9 @@ module executeBlockESpec (bs : BlockStore) (block : Block) where
   postulate -- TODO: prove
     contract : (isOk : Ok) → ContractOk (proj₁ isOk)
 
-module executeAndInsertBlockESpec (bs0 : BlockStore) (block : Block) where
+module executeAndInsertBlockESpec (bs0 : BlockStore) (vblock : ValidBlock) where
+  block   = vbBlock vblock
+  block-c = vbValid vblock
   open executeAndInsertBlockE bs0 block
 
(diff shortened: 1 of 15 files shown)


mark-moir:
Closing as this was merged together with #160.

mark-moir:
Superseded by #167

PR #161, opened 2021-09-20
prove EitherD/Either versions of init equivalent; init contract

haroldcarr:
prefix a LOT of stuff with `fake` to make it easier to identify and remove later

Impl/Handle
- prove EitherD/Either versions of initEMWithOutput are equivalent
- fake renames
- reordered stuff in file

Impl/Handle/InitProperties
- new file
- initial shot at contract

### DIFF
--- a/LibraBFT/Impl/Handle.agda
+++ b/LibraBFT/Impl/Handle.agda
@@ -72,7 +72,7 @@ postulate -- TODO-1: initPE, initBS, initRS
 initRM : RoundManager
 initRM = RoundManager∙new
            ObmNeedFetch∙new
-           (EpochState∙new 1 (initVV genesisInfo))
+           (EpochState∙new 1 (initVV fakeGenesisInfo))
            initBS initRS initPE initPG initSR false
 
 -- Eventually, the initialization should establish properties we care about.
@@ -106,7 +106,7 @@ peerStep nid msg st = runHandler st (handle nid msg 0)
 
 InitAndHandlers : SystemInitAndHandlers ℓ-RoundManager ConcSysParms
 InitAndHandlers = mkSysInitAndHandlers
-                    genesisInfo
+                    fakeGenesisInfo
                     initRM
                     initWrapper
                     peerStep
--- a/LibraBFT/Impl/Handle.agda
+++ b/LibraBFT/Impl/Handle.agda
@@ -63,7 +63,7 @@ initSR =
 initPG : ProposalGenerator
 initPG = ProposalGenerator∙new 0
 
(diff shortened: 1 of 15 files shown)


LibraBFT/Impl/Handle.agda
```diff
@@ -123,10 +104,52 @@ runHandler st handler = ×-map₂ (outputsToActions {st}) (proj₂ (LBFT-run han
...
+initEMWithOutput≡ {x} iewo
+  with GenKeyFile.create 1 (0 ∷ 1 ∷ 2 ∷ 3 ∷ [])
+... | Right (nf , _ , vss , vv , pe , liws)
+  with Init.initialize 0 (nf , liws , vss , vv , pe) now ObmNeedFetch∙new pg
+... | Right y rewrite iewo = cong Right refl
+
+------------------------------------------------------------------------------
+-- TODO : ASK CHRIS : regarding EitherD-run
+
+zzz : EitherD ErrLog ℕ
```
mark-moir: Did you check in the right version of this?  This seems to be still in the state right after we figured out that we were confused over the difference between `initEMWithOutput ≡ Right x  and `EitherD-run initEMWithOutput ≡ Right x`.  We have one post-confusion case proved above, but not the 'Left` case, which should be equally trivial.  Also, I think we want (at least) the other direction...

Actually, I just pushed a commit to make clearer what I was about to say.  It shows that the proof that the two functions are equivalent is trivial, as expected/hoped.  All we have to do is the case analysis and rewrite/refl does the rest.

haroldcarr: Oh yeah - thanks for the update.

LibraBFT/Impl/Handle/InitProperties.agda
```diff
@@ -0,0 +1,38 @@
...
+    field
+       rmInv : Σ RoundManager
+                 λ rm → em ^∙ emProcessor ≡ just (RoundProcessorNormal rm)
+                      × Util.Invariants.RoundManagerInv rm
+
+  Contract : Either ErrLog (EpochManager × List Output) → Set
+  Contract (Left _)          = ⊤
+  Contract (Right (em , lo)) = ContractOk em lo
+
+--  open initEMWithOutputSpec
```
mark-moir, line 30: Might as well uncomment the postulated proof :smile:.

mark-moir (approved):
Sure, gahead :smile:.

LibraBFT/Impl/Handle/InitProperties.agda
haroldcarr: It is not ready to be uncommented.

PR #163, opened 2021-09-21
Yasm/Base init return Maybe; rename Genesis to Bootstrap

haroldcarr:
Yasm/Base
- change init to return Maybe
- rename init to bootstrap

Rename genesis-like names to bootstrap* names.

### DIFF
--- a/LibraBFT/Yasm/Base.agda
+++ b/LibraBFT/Yasm/Base.agda
@@ -17,9 +17,9 @@ module LibraBFT.Yasm.Base (ℓ-PeerState : Level) where
   field
     PeerId    : Set
     _≟PeerId_ : ∀ (p₁ p₂ : PeerId) → Dec (p₁ ≡ p₂)
-    Genesis   : Set
+    Bootstrap : Set
     -- A relation specifying what Signatures are included in genInfo
-    ∈GenInfo  : Genesis → Signature → Set
+    ∈BootstrapInfo  : Bootstrap → Signature → Set
     PeerState : Set ℓ-PeerState
     Msg       : Set
     Part      : Set -- Types of interest that can be represented in Msgs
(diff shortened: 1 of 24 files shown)


LibraBFT/Yasm/Base.agda
```diff
@@ -48,9 +48,11 @@ module LibraBFT.Yasm.Base (ℓ-PeerState : Level) where
...
 
-      -- Initializes a potentially-empty state with an EpochConfig
-      init : PeerId → Genesis → PeerState × List (Action Msg)
+      -- Bootstraps a peer.
+      -- TODO-1 : replace Maybe with Either after adding error type (to mkSysTypeParms ?).
+      bootstrap : PeerId → Genesis → Maybe (PeerState × List (Action Msg))
 
       -- Handles a message on a previously initialized peer.
-      handle : PeerId → Msg → PeerState → PeerState × List (Action Msg)
+      -- TODO-1 : replace Maybe with Either after adding error type (to mkSysTypeParms ?).
```
mark-moir, line 56: I don't think it makes sense to make `handle` return a `Maybe`.  In case a peer handler encounters an error, it can decide to update the peer state or not (note that it already has one), and it can log errors via the `List (Action Msg)`.  We need a `Maybe` for bootstrapping above, because the `bootstrap` function might be unable to create a peer state at all, e.g., if `Genesis` information is invalid.

LibraBFT/Yasm/Base.agda
```diff
@@ -48,9 +48,11 @@ module LibraBFT.Yasm.Base (ℓ-PeerState : Level) where
       -- Represents an uninitialised PeerState, about which we know nothing whatsoever
       initPS    : PeerState
 
-      -- Initializes a potentially-empty state with an EpochConfig
-      init : PeerId → Genesis → PeerState × List (Action Msg)
+      -- Bootstraps a peer.
+      -- TODO-1 : replace Maybe with Either after adding error type (to mkSysTypeParms ?).
+      bootstrap : PeerId → Genesis → Maybe (PeerState × List (Action Msg))
```
mark-moir, line 53: I thought we decided to rename `Genesis` to (something like) `BootstrapInfo`, as `Genesis` sounds a bit application-centric and therefore doesn't belong in the system model.  I'm ambivalent regarding `init` vs. `bootstrap` for the function name.

mark-moir (changes requested):
@haroldcarr see comments.  I believe the most disruptive change here is actually unnecessary, which is good news :smile:.  Thanks.

mark-moir:
P.S.  Love the spelling of the title though!

LibraBFT/Yasm/Base.agda
haroldcarr: One thing at a time.   First step is `Maybe` on `bootstrap`.

haroldcarr:
Ha! "Yasm/Bass" - I guess my fingers have a mind of their own!

haroldcarr:
@mark-moir This is now an official PR.
I am not going to make any more changes in this branch, other than those requested by review.

mark-moir:
Well THAT was fun :smile:.  APPROVED.  Thanks.

PR #164, opened 2021-09-21
Proved α-EC-VV

lisandrasilva:
Proved `α-EC-VV`:  given a `ValidatorVerifierCorrect`, that is, a well-formed set of definitions that defines an `EpochConfig`, including the BFT-assumption, `α-EC-VV` will compute this `EpochConfig` by abstracting away the unnecessary pieces from `ValidatorVerifier`. Note that the BFT-assumption is part of the `ValidatorVerifierCorrect`, therefore in order to instantiate it the user will have to postulate the BFT-assumption whenever necessary.


mark-moir (approved):
Looks great @lisandrasilva, thanks!  My review comments are minor (comments, formatting, etc., so I am going to Just Do Them after merging to avoid the round trip.  Thanks again.

PR #168, opened 2021-09-28
Constrain `step-init` to occur only if init handler is successul

mark-moir:
As we prepare to plug in a real initialisation handler, we must account for the possibility that initialisation fails.  This PR constrains `step-init` to occur only if the handler is successful.  It also:
* Reverts unnecessary changes to make `StepPeerState` have a `Maybe`, which was unnecessary because even unitialised peers have default peer states.
* includes notes about a more general approach that would allow for a failed initialisation to have side effects such as logging or sending messages; not a priority for now and it will be more disruptive to existing proofs
* jettisons `ImplFake.*` which has served its purpose well but now is a burden to maintain
* jettisons `PendingVotesSpec`, which is unnecessary because we are proving the property that it tests
* "fixes" proofs broken by the change; in a number of cases the "fix" is `obm-dangerous-magic` with a note about what the `Contract` for the real initialisation handler will need to provide to enable the proof to be fixed
* introduces cases that Agda could previously eliminate because it knew the fake init handler sends no messages; similarly commented on what is needed from the `Contract` to discharge those proofs


haroldcarr:
I see that the main changes, in terms of init, are:

- Yasm.System
  - StepPeerState.step-init
    - takes proofs of uninitialized and that bootstrap succeeds
- Impl.Handle.Properties
  - updated accordingly
  - uses obm-dangerous-magic as placeholders for where I need to be working

I am going to
- merge this branch into main
- rebase my branch with main
- continue working

PR #174, opened 2021-10-19
Using init handler contract to prove more step-init properties

haroldcarr:


LibraBFT/Impl/Properties/VotesOnce.agda
```diff
@@ -40,21 +40,31 @@ open import LibraBFT.Yasm.Yasm ℓ-RoundManager ℓ-VSFP ConcSysParms
...
+  → ReachableSystemState pre
+  → StepPeerState pid (msgPool pre) (initialised pre) (peerStates pre pid) (s' , acts)
+  → v ⊂Msg m
+  → send m ∈ acts
+  → (sig : WithVerSig pk v)
+  → Meta-Honest-PK pk
+  → ¬ (∈BootstrapInfo-impl fakeBootstrapInfo (ver-signature sig))
+  → ¬ MsgWithSig∈ pk (ver-signature sig) (msgPool pre)
+  → LastVoteIs s' v
+newVote⇒lv≡ {pre}{pid}{s'}{v = v}{m}{pk} preach (step-init handler-pid-bsi≡just-rm×acts uni) v⊂m send∈acts sig hpk ¬bootstrap ¬mws∈pool
```
mark-moir, line 60: AFAICT, `pid` is the only one of these implicit parameters that is actually used, so you could just have `{pid = pid}` instead of listing them all.  Not a big deal, but it does make it more readable if unnecessary ones are omitted.  Similar comments in other cases below.

LibraBFT/Impl/Properties/VotesOnce.agda
```diff
@@ -40,21 +40,31 @@ open import LibraBFT.Yasm.Yasm ℓ-RoundManager ℓ-VSFP ConcSysParms
...
+  with IP.initHandlerSpec.contract pid fakeBootstrapInfo handler-pid-bsi≡just-rm×acts
```
mark-moir, line 61: I understand why you use these long variable names while developing, but I think they do more harm than good for readability, so I'd prefer you use shorter names that capture the important intuition but don't spell out so much detail.  Up to you do decide whether to change them.

mark-moir (approved):
Well done!  A few minor comments.  When you have addressed them (or not) to your satisfaction, APPROVED.  Thanks @haroldcarr.

PR #178, opened 2021-11-17
added more documentation to EpochManager processMessage and obmStartloop

haroldcarr:


LibraBFT/Impl/Consensus/EpochManager.agda
```diff
@@ -423,6 +440,14 @@ start self0 now obmPayload obmNeedFetch obmProposalGenerator obmLedgerInfoWithSi
   - in other cases it sends messages (via 'stps') to other nodes
 
 The following "implementation" compiles, but has lots of TEMPORARY things to get it to compile.
+In the proofs, the System Model "handles" input and output.
+Therefore 'obmStartLoop' will not be implemented nor "proved".
+It is here to show what is done in the Haskell code.
+If the Agda code were ever to be extracted to executable Haskell code,
+then this code would be completed.
+
+The system model will need to prove things about 'EpochManager.processMessage'.
```
mark-moir, line 449: This comment doesn't make sense.  The system model doesn't even know about `EpochManager.processMessage`, let alone prove anything about it.  I /think/ what you mean is that the system model will be instantiated with (something like) 'EpochManager.processMessage' and that will be the system about which we prove our properties.  Maybe?

LibraBFT/Impl/Consensus/EpochManager.agda
```diff
@@ -485,11 +510,11 @@ obmStartLoop self initializationOutput
         --(rm'' , to'') ← DAR.runOutputHandler rm' to pe o  oh
         rm'' ← pure rm -- TEMPORARY for previous two lines
         loop (setProcessor em rm'') {-to''-} rlec
-      (PMNewEpochManager em' newEpochInitializationOutput) → do
+      (PMNewEpoch em' newEpochInitializationOutput) → do
        eitherSD (em' ^∙ emObmRoundManager) ee $ λ rm → do
         -- (rm', to') ← DAR.runOutputHandler   rm  to pe newEpochInitializationOutput oh
         rm' ← pure rm -- TEMPORARY for previous line
-        loop (setProcessor em' rm') {-to'-} rlec -- TODO Set₁ != Set
+        loop (setProcessor em' rm') {-to'-} RSNothing -- IMPORTANT*** : reset RLEC state
```
mark-moir, line 517: Yes, it is important, but then it's important to get _many_ things right :smile:.  So, I would keep the comment but remove "IMPORTANT***".  Great catch, btw!

mark-moir (approved):
@haroldcarr LGTM, approved when you've addressed my minor comments.  Thanks.

LibraBFT/Impl/Consensus/EpochManager.agda
haroldcarr: Yes - "instantiate".  DONE

LibraBFT/Impl/Consensus/EpochManager.agda
haroldcarr: DONE

PR #179, opened 2021-12-16
refactor non-LibraBFT into non-LibraBFT directories/modules

haroldcarr:
All `*.agda` code now lives is `src`.

All code that is  independent of `src/LibraBFT` now lives in
- `src/Dijkstra`
- `src/Haskell`
- `src/Optics`
- `src/Util`

The only files that needed to be changed due to `src` are:
- `LibraBFT.agda-lib` - so it points to `src`
- `Scripts/gen-everything.sh` - so it knows about `src` (and so it really does gen EVERYTHING)
- `Scripts/run-everything.sh` - so it knows about `src`
-  `src/Optics/Example.agda` - previously, it was not included in `Everything.agda` and thus has not been checked.   Now it is included and initially did not compile.   I fixed the bugs.

The `src/Haskell` code has been further subdivided into `Modules`.


mark-moir:
Thanks @haroldcarr.  LLIGSTM (looks like a good start to me :smile:).  It seems to me that you have not gone far enough, though.  I believe that everything you currently have under `LibraBFT.ImplShared.Util.Dijkstra` is also completely independent of LBFT and therefore should also be moved out of the `LibraBFT` name space.

While the paper we're currently working on is not going to talk about *proofs* using Dijkstra weakest precondition machinery, we are certainly talking about the setup for it.  For example, we're talking about stuff under `LibraBFT.ImplShared.Util.Dijkstra.Syntax` in the paper.

We could consider whether to include this stuff under a separate directory (e.g., `Dijkstra` that lives next to `Optics` and `Haskell`) or put it under `Haskell`.  I think I lean towards a separate `Dijkstra` directory.  That raises the question of the where the additional constructors for branching combinators (e.g., `RWS-if`) should live, given that the sole motivation for them is for interfacing with proofs.  However, I guess someone could use them with a different proof "back end" other than Dijkstra wp, so maybe it's fine that the additional constructors live under Haskell.

Thoughts?  Thanks.

mark-moir:
Another suggestion is to further refine/organize things under the `Haskell` namespace.  For example, we could have `Haskell.Monad` and `Haskell.Either`; the latter would be analogous to (though much simpler than) `Haskell.RWS`.  Finally, there are various things that are included because Agda's library doesn't include everything provided by Haskell libraries.  It would be nice to organise so that these are together, rather than intermingled with other things that don't fit that description.

Note: We had talked about establishing a structure following that of another repo.  Not sure to what extent you have done that already, how compatible it is with the suggestions above, etc.  But I'd prefer to give up on mimicking the structure of the other repo if it saves time or eliminates some difficulty that it entails.

haroldcarr:
Agree on moving `Dijkstra` out from under `LibraBFT`.
For the 1st pass I'll put it parallel to `Optics` and `Haskell`.

As for further structure under `Haskell` - I had thought that maybe that might be a next step.
That said, there is so little in `Haskell.Prelude` at the moment that I am not sure that moving it into separate files is that useful right now.   One concern:   In Haskell, the more files one has in a project, the longer it takes to compile.  I've noticed the same thing with Agda.

I think perhaps, as a 1st pass, I can organize the contents of `Haskell.Prelude` with internal modules that when can move into separate files in the future if there is some benefit.  Or, instead of modules, at least clearly marked (with --------------------) areas of the file.   Thoughts?

mark-moir:
> Agree on moving `Dijkstra` out from under `LibraBFT`. For the 1st pass I'll put it parallel to `Optics` and `Haskell`.
> 
> As for further structure under `Haskell` - I had thought that maybe that might be a next step. That said, there is so little in `Haskell.Prelude` at the moment that I am not sure that moving it into separate files is that useful right now. One concern: In Haskell, the more files one has in a project, the longer it takes to compile. I've noticed the same thing with Agda.
> 
> I think perhaps, as a 1st pass, I can organize the contents of `Haskell.Prelude` with internal modules that when can move into separate files in the future if there is some benefit. Or, instead of modules, at least clearly marked (with --------------------) areas of the file. Thoughts?

Thanks @haroldcarr.  My experience/understanding is that main "compile time" issues we face are with larger files, and we keep saying we should break some of them up.  In any case, what we're discussing here is pretty small, so I think we should favour the organisation that makes most sense, and we'll deal with any performance issues when/if they arise.

I do like the idea of making the `Either` and `RWS` "parallel", because it makes it easy to first understand the much simpler `Either` and then follow the parallel structure in `RWS`.

haroldcarr:
Regarding compiletime : I meant that `./Scripts/run-everything xxx` takes longer when there are more files.  But, yeah, "loading" smaller files into Agda in Emacs is faster.

haroldcarr:
@mark-moir Thanks.  Back to you.

mark-moir:
Looks good @haroldcarr thanks.  However, I am curious if there is some rationale for `RWS*` not being under `Modules` or if this is just an oversight.

haroldcarr:
Thanks @mark-moir .    I agree that it should become `Haskell.Modules.RWS`
Should I do it now?

mark-moir:
If that's how you feel like spending your Sunday afternoon, sure :smile:.  In case you didn't notice, I just updated my comment (assuming you were spending your Sunday afternoon more sensibly and wouldn't notice!)

haroldcarr:
I happened to be passing by my computer and saw the notice.
This should not take long.

mark-moir:
> I happened to be passing by my computer and saw the notice. This should not take long.

Cool.  Forgot to say: nice cleanup putting stuff under `src` and good catch that `run-everything.sh` was misspelled (and I prefer your fix over a spelling change :smile:).

haroldcarr:
Now back to our regularly scheduled Sunday program.  ;-)

PR #181, opened 2022-01-12
moved Yasm to top level

haroldcarr:
moved `Yasm` to top level; moved everything in `LibraBFT` it depended on to `Util`

Note: the things moved to `Util` can probably use some refactoring too, but that should be done in a separate PR.


mark-moir:
@haroldcarr, mostly looks good.  I agree that a) `Util` needs some refactoring and b) this can be done in a future pull request.

That said, I think it would be better to (mostly) revert the move of `LibraBFT.Util.Types` because its contents are specific to our `LibraBFT` use case (but perhaps `LibraBFT.Base.Types` is more appropriate than `LibraBFT.Util.Types`?).

(Note that `Yasm` no longer depends on `Epoch` and `Round`.)

`EPRound` and everything below are unused at present but may be useful in future for cross-epoch correctness conditions, so might as well keep them (also in `LibraBFT.Base.Types`?).

`Meta` is unused at present, and is not use case-specific.  I think it makes sense to keep it nonetheless to remind us of the issue that it was intended to address, namely that we could inadvertently make the implementation use "meta" information that is not available to peers (e.g., which `PK` are honest, which we currently avoid only via naming convention).  It's not use case specific but on the other hand, given that we haven't used it successfully, documented it, etc. perhaps it should remain under `LibraBFT`.  Happy to accept your judgment, after which APPROVED.

Thanks.

Cheers

Mark

haroldcarr:
My mistake - I meant to move that back.
Good catch!
DONE.

PR #182, opened 2022-01-13
Update and expand `PeerHandlerContracts.org`, discuss `EitherD` approaches

mark-moir:
This PR expands on the documentation of how we approach proving properties about peer handlers.  It reorganises the document somewhat, so that all of the weakest precondition machinery and discussion of its use is in terms of `RWS`, 
leaving `EitherD` to be explained later with reference to it.

Hopefully it is mostly OK.  However, I have left a `TODO` in the section about `EitherD`, because I think we should discuss how we can (and whether we should/will) streamline this stuff.  I am thinking along the lines of:
* pragmatically accept that we will usually write peer code directly in the `EitherD` monad
* use `EitherD-run` whenever we need to pattern match on results (perhaps with some syntactic sugar to make it less intrusive)
* for any cases where the resulting code is not sufficiently obviously equivalent to the relevant Haskell code, write the `Either` version and prove that it is "equivalent" to the corresponding `EitherD` version about which we prove properties.


mark-moir:
This is superseded by #184.  Closing.

PR #184, opened 2022-01-25
Simplify and document approach to dealing with monadic code in `Either`, other documentation updates

mark-moir:
In this PR, we refine the approach we use to representing and reasoning about code written in the `Either` monad in Haskell, so that we write in `EitherD` and create an `Either` variant using `toEither`, thus eliminating the previous recommendation to have two different `EitherD` variants.   We also update some examples and add some "tutorial" comments.

Also, the documentation is updated and reorganised, so that everything not related directly to `EitherD` comes first, and `EitherD` is discussed last.


mark-moir:
This was superseded by #187, and lived way too long thereafter.  Closing!

PR #187, opened 2022-02-10
Refine definitions of `Either` variants, proofs, prove that `EitherD-weakestPre` really is weakest, and update documentation

mark-moir:
* Prove that `EitherD-weakestPre m P` really is the weakest precondition required to ensure that predicate `P` holds for the result of running `EitherD` program `m`.  This is shown by proving that, for any `m : EitherD E A`, if `P (EitherD-run m)` holds, then `EitherD-weakestPre m P` holds.  This is stated as property `Post⇒wp`, and proved in `EitherD-wp-is-weakest`.

* Add an "original" version of `ensureRoundAndSyncUpM`, which looks almost identical to the original Haskell code, and provide a (trivial) proof that it is equivalent the the version that is broken into steps for proofs.  We do not intend to do so in all cases, because the correspondence is usually straightforward.  This demonstrates that we can do so when needed.

* Similarly for `insertBlockE`

* Update definitions of the following functions to be defined first as `EitherD`, then an `Either` variant derived using `toEither`, along with equivalent definitions in abstract blocks and lemmas for rewriting the abstract variant in order to reveal the details of the `EitherD` variant when needed.

  * `BlockTree.addChild`
  * `BlockTree.insertBlockE`
  * `BlockTree.insertQuorumCertE`
  * `BlockStore.executeAndInsertBlockE`
  * `BlockStore.executeBlockE`

* Reformulate `executeBlockE` contract, prove it, fix proofs that use it

### DIFF
--- a/src/Dijkstra/EitherD.agda
+++ b/src/Dijkstra/EitherD.agda
@@ -81,6 +81,9 @@ EitherD-weakestPre-bindPost f P (Left x) =
 EitherD-weakestPre-bindPost f P (Right y) =
   ∀ c → c ≡ y → EitherD-weakestPre (f c) P
 
+-- For any post contition P and EitherD program m, if
+-- EitherD-weakestPre m P holds, then P holds of the result
+-- of running m.
 EitherD-Contract : (m : EitherD E A) → Set₁
 EitherD-Contract{E}{A} m =
   (P : EitherD-Post E A)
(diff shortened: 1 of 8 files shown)


src/Dijkstra/EitherD.agda
```diff
@@ -113,6 +116,43 @@ EitherD-contract (EitherD-maybe f₁ f₂ nothing) P wp =
 EitherD-contract (EitherD-maybe f₁ f₂ (just x)) P wp =
   EitherD-contract (f₂ x) P (proj₂ wp x refl)
 
+-- If P holds of the result of running m, then EitherD-weakestPre m P
+-- holds.  This shows that EitherD-weakestPre really does compute the
+-- weakest precondition, which is interesting and comforting, but not
+-- important for the correctness of results proved using it.
+-- Chris said: EitherD-weakestPre m P is a terminal object in the
+-- (provability) category of preconditions for P
+-- Mark agreed :-)
```
cwjnkins, line 125: Proof: by intimidation

PR #188, opened 2022-03-25
Add summary and links for documentation and papers

mark-moir:


mark-moir:
Self approving.  To be updated with links when finalised.

PR #189, opened 2022-04-26
Generalization of EitherD, RWS

cwjnkins:
This PR contains a POC generalization of the repo's current approach to program verification. A single datatype, `AST`, can be instantiated to a type equivalent to the existing `RWS` or to one equivalent to `EitherD`. Furthermore, the work of adding predicate transformers for branching operations need not be repeated for each such instance (as is done in the code currently), but can be generically added to a set of `AST` operations.


cwjnkins:
Perhaps this can merged into its own branch for now, rather than main, since it is still under development?

PR #192, opened 2022-05-03
Demonstrate where I'm getting blocked, with a couple of different approaches

mark-moir:
This PR demonstrates the difficulty that this particular mortal keeps running into, no matter how he attempts to approach proving the `divWorks` property.  Maybe a higher being will figure out what I'm missing and together we can come up with useful guidance for other mortals :smile:.

**Update**: I made some progress.  First, I cleaned up the cases and dispatched some easy ones.  Then, I managed to  make progress where I was stuck so many times.  Two incredibly minor details were preventing progress:

1. I had not done `open ASTPredTrans MaybePT` so Agda was not expanding `bindPT`, so was not presenting the goal I needed.
2. The definition of `bindPT` was for `ASTPredTrans.bindPT   MaybePT f unit Post x`, rather than `ASTPredTrans.bindPT   MaybePT f i Post x`.  Even though there is only one constructor for `Unit` (and therefore no missing cases reported), Agda did not figure out the this was the case for a general `i`, and again did not present the goal I wanted.

Good to make progress, but frustrating to have spent so much time due to these minor details!

### DIFF
--- a/src/Dijkstra/AST/Core.agda
+++ b/src/Dijkstra/AST/Core.agda
@@ -47,17 +47,17 @@ record ASTTypes : Set₁ where
   PredTrans : (A : Set) → Set₁
   PredTrans A = (P : Post A) → Pre
 
-  _⊆ᵢ_ : (P₁ P₂ : Pre) → Set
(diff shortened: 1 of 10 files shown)


mark-moir:
This has served its purpose, so now I closing

PR #193, opened 2022-05-03
Dijkstra: AST: Maybe: proof that div works

cwjnkins:
We can discuss this solution in the meeting today, if desired.

Signed-off-by: Christa Jenkins <chris.w.jenkins@oracle.com>


mark-moir:
@cwjnkins  Thanks Christa!  This is an important and useful example that people who use (y)our framework should internalize.

PR #194, opened 2022-05-03
Add ASTs for `Either` and `Maybe`, along with examples and proofs

mark-moir:
This combines work done by @mark-moir, @haroldcarr and @cwjnkins to add `Either` and `Maybe` AST versions and prove required properties plus some examples.  I'm proposing to merge this to branch `chris-pr-189`, where we can continue this work before merging to `main`.


haroldcarr (approved):
I am considering the time we have already spent working together on this as the review, so am approving without looking in detail at this moment.

PR #197, opened 2022-05-05
Minor tweaks and comments on `AST` framework and use cases (`Either` and `Maybe`)

mark-moir:
Signed-off-by: Mark Moir <mark.moir@oracle.com>

PR #198, opened 2022-05-05
Create `AST` version of `insertBlockE`, prove its contract

mark-moir:
This PR provides the first experience with using the new framework to prove something about LBFT. Specifically, it adds an `AST` version of `insertBlockE`, and proves its contract (after refining it to require an additional assumption to be passed in, which also required some changes to the proofs "up the stack" to feed this assumption).

This PR builds on [#197](https://github.com/oracle/bft-consensus-agda/pull/197), which should be merged first.


mark-moir:
@haroldcarr this one has been sitting while.  I've updated it and resolved conflicts.  Do you want to have a quick look before I merge it?  Thanks.

mark-moir:
@haroldcarr I'm merging this to simplify the next one.  We can still do the live code review we planned and make any changes in a separate PR.

PR #208, opened 2022-05-25
Some small fixes (level-generic `maybe`, convenient names for some `Dijkstra/AST/RWS.agda` functions)

cwjnkins:


haroldcarr (approved):
The aliases make sense and is totally understandable.

Adding levels to Maybe makes sense too, but I don't know enough to have figured out that was the problem with the example. 

Thanks!
APPROVED

mark-moir:
I approve this too.  Not sure that @cwjnkins can merge it, so I will do so now.

PR #218, opened 2022-06-08
added RWS example where framework really does do 90% of the work

haroldcarr:


src/Dijkstra/AST/Examples/RWS.agda
```diff
@@ -0,0 +1,51 @@
...
+open import Relation.Binary.PropositionalEquality
+
+module Example1 (A : Set) where
+
+  open import Data.Nat         renaming (ℕ to Nat) using (_+_ ; suc ; zero)
+  open import Dijkstra.AST.RWS A A (List A)
+  open        ASTPredTrans     RWSPT
+  open        RWSSyntax
+
+  sucn≡n+1 : ∀ (n : Nat) -> suc n ≡ n + 1
```
mark-moir, line 16: I think this is just `+-comm` from the standard library.

src/Dijkstra/AST/Examples/RWS.agda
```diff
@@ -0,0 +1,51 @@
...
+
+  ProgPost : (A × List A) -> (List A × List A × List A) -> Set
+  ProgPost (_ , si) (a , so , w) = length  a ≡ length so + 1
+                                 × length so ≡ length si + 1
+                                 × length  w ≡ 3
+
+  progPost : ∀ i -> ProgPost i (runRWS prog i)
+  progPost (e , s) with runRWS prog (e , s)
+  ... | (a , st , wr)
+    rewrite sucn≡n+1 (foldr (λ _ → suc) 0 s)
```
mark-moir, line 39: Why not use `length` (from `Data.List`)?

src/Dijkstra/AST/Examples/RWS.agda
```diff
@@ -0,0 +1,51 @@
...
+    = refl , refl , refl
+
+  progPostWP : ∀ i -> predTrans prog (ProgPost i) i
```
mark-moir, line 42: It would be useful to add a little commentary about what's going on here.  I think the reason this one goes so smoothly, relative to the `Maybe` and `Either` examples we've been working on, is that the latter types are sums, so we need to break the proof obligations down into cases, whereas for `RWS` there is no such challenge.

src/Dijkstra/AST/Examples/RWS.agda
```diff
@@ -0,0 +1,51 @@
+module Dijkstra.AST.Examples.RWS where
```
mark-moir, line 1: Please add the copyright notice.

src/Dijkstra/AST/Examples/RWS.agda
```diff
@@ -0,0 +1,51 @@
...
+
+open import Data.Product      using (_×_ ; _,_)
+open import Dijkstra.AST.Core
+open import Function          using (case_of_)
```
mark-moir, line 5: Not used?

mark-moir (approved):
APPROVED modulo review comments.

src/Dijkstra/AST/Examples/RWS.agda
haroldcarr: REMOVED.

src/Dijkstra/AST/Examples/RWS.agda
haroldcarr: DONE.

src/Dijkstra/AST/Examples/RWS.agda
haroldcarr: `+-comm` is a property about two Nats.   This property I defined is about one Nat.
I couldn't find anything in `Data.Nat.Properties` like it (to my surprise).

src/Dijkstra/AST/Examples/RWS.agda
haroldcarr: DONE

src/Dijkstra/AST/Examples/RWS.agda
mark-moir: Apologies for response in the form of a PR on a branch called `mark-tweak-rws-example-and-harold` :smile:.  #221

PR #219, opened 2022-06-08
either cargo cult maybe

haroldcarr:
rename `runEither` to `runEitherAST` (for consistency with `Maybe`)

implement `bindCont`

(next, in a future PR :  `eitherPTBindLemma`)


mark-moir:
@haroldcarr if you don't mind, I'll reject this one.  The reason is that I'm working on more disruptive changes to make it nicer for the user, which also does the renaming you've done here.  I also think that `bindCont` can be generalised so we don't need it for each monad, which I had planned to do.

PR #220, opened 2022-06-09
Make extended `AST` versions default, streamline their use, update examples and proofs

mark-moir:
The purpose of this PR is to make it easier to use the framework, including branching operations.
Rather than considering the branching operations as "extensions" to an underlying ("base") AST, we
consider the version extended with branching operation as the first target, even for programs that
don't use branching operations (initially).  We then streamline the importing and opening of the
extended AST and associated properties, as demonstrated in the updated examples.  Finally, the
BlockTree definitions and proofs are updated to use the new style, demonstrating the simpler
importing/opening facilitated by this PR.


haroldcarr (approved):
approved --- based on live code review

mark-moir:
> approved --- based on live code review

Thanks, changes made during/after live review have been done.

PR #225, opened 2022-06-16
make RWSBase completely public

haroldcarr:


mark-moir:
You took this a little further than we discussed, and reflecting further on this makes me think we shoyld actually go even less far than we discussed :smile:.

In general, a user would use `RWSAST` and should be discouraged from using `RWSASTBase`, as there's no benefit to using the unextended version, and there is a disadvantage to using it because one might subsequently want to use the branching operations, which would require changing from `RWSASTBase` to `RWSAST`.

I know your motivation is to enable some examples illustrating equivalent programs expressed using low-level `RWSCmd` constructors, whereas "normal" use would be in terms of the extended `AST` and would use `RWSSyntax`, rather than low-level constructors.  I think it makes sense to expose `RWSbindPost`, `RWSpassPost` and `RWSlistenPost` because they may be used in proofs, but otherwise details of `RWSASTBase` should not be exposed by default.

I think use cases such as yours that do want to use the lower level constructors can explicitly open `RWSBase.RWSCmd` themselves, rather than this being the default via public `open`s in `RWSAST`.

haroldcarr:
@mark-moir 
"further than we discussed" --- what I added in Zoom was `open RWSBase public`.
But later I saw the other selective open.   So that's why I went with that.

Now I understand:
- most users would use RWS (that includes RWSAST and RWSSyntax)
- do an explicit `open RWSBase.RWSCmd` to talk about constructor when needed.
  - I updated my local example to do that.

PR #238, opened 2022-06-21
First cut making `Dijkstra.AST.*` independent of other namespaces

mark-moir:
An initial attempt to make `Dijkstra.AST.*` independent of other namespaces, particularly `Haskell.*`.  Everything under `Dijkstra.AST` compiles, but `LibraBFT.Impl.Consensus.BlockStorage.BlockTree` does not.  We have some challenges where we want to mix Haskell and AST that I have not ironed out yet.  In some cases I have duplicated definitions in `Dijkstra.AST.Prelude` and friends, but this has led to some difficulties around things like where `Monad` instances come from, and having the right syntax definitions in the right context.  Need to step back and think more about this, but sharing what I have for now.

Signed-off-by: Mark Moir <mark.moir@oracle.com>


src/Dijkstra/AST/Branching.agda
```diff
@@ -3,29 +3,20 @@
    Copyright (c) 2022, Oracle and/or its affiliates.
    Licensed under the Universal Permissive License v 1.0 as shown at https://opensource.oracle.com/licenses/upl
 -}
-
```
haroldcarr, line 6: Put the blank line between copyright and `module` back in.
It's easier for users to scan that way.

src/Dijkstra/AST/Branching.agda
```diff
@@ -3,29 +3,20 @@
...
-open import Data.Fin
-open import Data.Product using (_×_ ; _,_ ; proj₁ ; proj₂)
-open import Data.Unit
 open import Dijkstra.AST.Core
-open import Function
-open import Haskell.Prelude
-import      Level
-import      Level.Literals as Level using (#_)
-open import Relation.Binary.PropositionalEquality
+open import Dijkstra.AST.Prelude
```
haroldcarr, line 9: Good idea to put the "Dijkstra" needs into its own `Prelude`.

src/Dijkstra/AST/Core.agda
```diff
@@ -11,7 +11,6 @@ open import Data.Fin
 open import Data.Product using (_×_ ; _,_ ; proj₁ ; proj₂)
```
haroldcarr, line 11: curious, would it be good to have all `Dijkstra` modules import "Dijkstra"'s `Prelude` rather than individual stuff?

src/Dijkstra/AST/Examples/Maybe/Partiality.agda
```diff
@@ -4,15 +4,16 @@
    Licensed under the Universal Permissive License v 1.0 as shown at https://opensource.oracle.com/licenses/upl
 -}
 
-open import Agda.Builtin.Unit using (⊤; tt)
-open import Data.Empty        using (⊥; ⊥-elim)
+open import Dijkstra.AST.Prelude
+-- open import Agda.Builtin.Unit using (⊤; tt)
```
haroldcarr, line 8: Why are these (and below) commented out, rather than deleted?

src/Dijkstra/AST/Syntax.agda
```diff
@@ -6,15 +6,39 @@
...
 module Dijkstra.AST.Syntax where
 
-open import Data.Empty
-open import Data.Product using (_×_ ; _,_ ; proj₁ ; proj₂)
-open import Data.Unit
-open import Function
-open import Haskell.Prelude
+open import Level renaming (_⊔_ to _ℓ⊔_ ; suc to ℓ+1)
+
+record Functor  {ℓ₁ ℓ₂ : Level} (F : Set ℓ₁ → Set ℓ₂) : Set (ℓ₂ ℓ⊔ ℓ+1 ℓ₁) where
```
haroldcarr, line 11: Functor, Applicative and Monad should go into a separate file (like we did in under `Haskell`.

haroldcarr (commented):
looks good so far

mark-moir:
I am creating a new version of this PR, with the issues outlined above addressed, so closing this one.

PR #243, opened 2022-06-23
Make `Necessary` in the style of `Sufficient`, eliminate `predTrans-is-weakest`

mark-moir:
The PR, makes `Necessary` in the style of `Sufficient`, requiring the user to provide obligations for each `AST` constructor, and then proving necessity using them, and extending these to branching operations.  It also adapts existing examples to use it, retiring the more ad-hoc `predTrans-is-weakest`.


mark-moir:
Thanks for the live review @cwjnkins.  Merging after some more cleanups.

PR #245, opened 2022-06-28
Make `Dijkstra.AST` (almost) entirely independent of Haskell

mark-moir:
This PR makes `Dijkstra.AST.*` (almost) entirely independent of other namespaces, particularly `Haskell.*`.  The "almost" is because `Monad`, `Applicative`, `Functor` and related definitions still live in `Haskell.Modules.FunctorApplicativeMonad`.  We could establish yet another namespace for these, and import it from `Haskell` as well, if anyone thinks it's worthwhile, but I have not done so.

This PR also fixes up all existing examples, which now have much simpler/cleaner imports because we have now packaged up `ConditionalExtensions` to provide everything needed.

It also splits each of `LibraBFT.Impl.Consensus.BlockStorage.[Properties.]BlockTree` into two files, one for all `AST`  versions and one for all non-`AST` versions.  This simplifies imports because we don't have to accommodate the old and new styles in the same file.


src/Dijkstra/AST/Branching.agda
```diff
@@ -279,6 +269,9 @@ module BranchingSyntax (BaseOps : ASTOps) where
                                   ; (lift false) → e
                                   }
```
haroldcarr, line 271: I would rewrite this comment as:
```
'maybeAST' follows the same argument order 'maybe' in Agda's stdlib.
```

Do not mention "flip it below" here - since does "below" mean somewhere in the "maybeAST" or somewhere else.  Besides, it is redundant - the MaybeD instance says what it does.

A rule of comments is "don't repeat what the code does".

You might put in WHY it is flipped (not the fact that it is flipped).  But if you do, put it next to where flip is used, not in the description of "maybeAST"

src/Dijkstra/AST/Branching.agda
```diff
@@ -289,6 +282,14 @@ module BranchingSyntax (BaseOps : ASTOps) where
                              ; (lift (just a)) → fA a
                              }
 
+  instance
+    open import Dijkstra.AST.Syntax
+    open import Dijkstra.Syntax
+    open MonadMaybeD public
```
haroldcarr, line 288: Speaking of "MaybeD" - what is the status of "MaybeD"?

A search shows it is used in PeerHandlerContracts and BlockTree-AST.

I am thinking, in an ideal world, this "old" stuff should go into a separate file (or be deprecated/replaced entirely).

src/Dijkstra/AST/Examples/Maybe/Branching.agda
```diff
@@ -4,14 +4,15 @@
...
+open import Dijkstra.AST.Prelude
 open import Data.Nat
-import      Level
-open import Util.Prelude hiding (bail)
 
 module Dijkstra.AST.Examples.Maybe.Branching where
 
 module Example-if (n : ℕ) where
   open import Dijkstra.AST.Maybe
+  open import Haskell.Prelude hiding (return)
```
haroldcarr, line 14: I think you mentioned these are to get functor, applicative, monad---right?

If that is correct, I suggest a qualified import `as HaskellPrelude` and using that prefix where needed.    Yes, ugly and long on purpose, since we really don't want to depend on `Haskell.Prelude`.

src/Dijkstra/AST/Prelude.agda
```diff
@@ -0,0 +1,50 @@
+module Dijkstra.AST.Prelude where
+
+open import Agda.Builtin.Unit
+     public
+open import Data.Bool                   using (Bool; true; false; not)
```
haroldcarr, line 5: I see a pattern of "only using" is on same line, otherwise, if there are other things (like renaming), on a separate line.

I think, in this file, do "using" on a separate indented line.

src/LibraBFT/Impl/Consensus/BlockStorage/BlockTree-AST.agda
```diff
@@ -0,0 +1,103 @@
+{- Byzantine Fault Tolerant Consensus Verification in Agda, version 0.9.
```
haroldcarr, line 1: I am not reviewing this file:

- I think I did a light review of these changes when this code was contained in the `BlockTree.agda` file.

So it is just separated into its own file now, right?

haroldcarr (approved):
Approved.
I think we should (in a subsequent PR) separate the last pieces of Haskell imported by Dijkstra.AST into its own top level module (like Haskell).

src/LibraBFT/Impl/Consensus/BlockStorage/BlockTree-AST.agda
mark-moir: > Approved. I think we should (in a subsequent PR) separate the last pieces of Haskell imported by Dijkstra.AST into its own top level module (like Haskell).

We can do that.  The reason I didn't do it already was that I wanted to seek input on appropriate organisation and naming.  If you or @cwjnkins have any suggestions, maybe we can do it before the paper deadline, otherwise it can wait.

mark-moir: > I am not reviewing this file:
> 
> * I think I did a light review of these changes when this code was contained in the `BlockTree.agda` file.
> 
> So it is just separated into its own file now, right?

Yes, correct.

src/Dijkstra/AST/Branching.agda
mark-moir: DONE

src/Dijkstra/AST/Branching.agda
mark-moir: There may be some confusion here.  MaybeD :-) you searched on the wrong branch?  Or maybeD you're talking about `MonadMaybeD`?  The latter is only about syntax, enabling the use of `maybeSD`, `caseMD`, etc.  It's nothing to do with the differences between the "old" and "new" style of reasoning, and does not need to be updated or eliminated.

src/Dijkstra/AST/Examples/Maybe/Branching.agda
mark-moir: Actually this import is no longer necessary.  FIXED.  Good catch, thanks.

src/Dijkstra/AST/Prelude.agda
mark-moir: DONE

src/Dijkstra/Syntax.agda
```diff
@@ -63,10 +63,19 @@ module _ {ℓ₁ ℓ₂} {M : Set ℓ₁ → Set ℓ₂} where
...
     ⦃ monad ⦄ : Monad M
+    -- This is the standard order for maybe arguments in Haskell;
+    -- see maybeSD below for the argument order we prefer in some
+    -- contexts
     maybeD : ∀ {A B : Set ℓ₁} → M B → (A → M B) → Maybe A → M B
 
 open MonadMaybeD ⦃ ... ⦄ public
 
+-- A Dijkstra version of maybeS, implemented using the version in
+-- Dijkstra.Syntax which has traditional argument order
```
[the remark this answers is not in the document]
mark-moir: I changed this to:

```
-- A maybeD variant with different argument order, which is favoured in
-- some use cases
```

The "some use cases" refers to the style we use in the `LibraBFT` code.

PR #246, opened 2022-06-29
Formatting, comments, naming for consistency with paper. 

mark-moir:
This PR contains no functional change, just formatting, naming, etc. for consistency with the paper (some changes made in the paper as well).

I'm creating the PR against branch `mark-make-ast-independent-of-haskell` so it's easy to review before #245 is merged, but it should be retargeted to `main-ast` after #245 is merged.

### DIFF
--- a/src/Dijkstra/AST/Either.agda
+++ b/src/Dijkstra/AST/Either.agda
@@ -5,7 +5,7 @@
 -}
 
 -- As we are interested in using Either for error handling (and providing the Either-bail command
--- for this purpose), we call the Left type "Err"
+-- for this purpose), we call the left type "Err"
 module Dijkstra.AST.Either (Err : Set) where
 
 open import Dijkstra.AST.Prelude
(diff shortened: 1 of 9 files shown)


src/Dijkstra/AST/Either.agda
```diff
@@ -5,7 +5,7 @@
 -}
 
 -- As we are interested in using Either for error handling (and providing the Either-bail command
--- for this purpose), we call the Left type "Err"
+-- for this purpose), we call the left type "Err"
```
haroldcarr, line 8: `Haskell.Modules.Either` still has patterns for lowercase `left` and `right`.

It seems "Haskell" should be Haskell and only have upper case `Left` and `Right'

And Dijkstra should either use Agda inj1/2 (or lower case `left` and `right`)

src/Dijkstra/AST/RWS.agda
```diff
@@ -94,24 +94,24 @@ module RWSBase where
...
-  opPT     RWSPT (RWSlocal l)         f P (ev , st) =
-    ∀ ev' → ev' ≡ l ev → f (lift unit) P (ev' , st)
-  opPT     RWSPT (RWStell out refl)   f P (ev , st) =
-    P (unit , st , out)
-  opPT     RWSPT (RWSlisten{A'} refl) f P (ev , st) =
-    f (lift unit) (RWSlistenPost P) (ev , st)
-  opPT     RWSPT{A} RWSpass           f P (ev , st) =
-    f (lift unit) (RWSpassPost P) (ev , st)
+  returnPT RWSPT x P (e , s) =
+    P (x , s , [])
```
haroldcarr, line 98: align with lines 102, 104, 106, ...

haroldcarr (approved):
APPROVED.

src/Dijkstra/AST/Either.agda
mark-moir: Good catch, thanks.  FIXED.

src/Dijkstra/AST/RWS.agda
mark-moir: You are the king of spotting inconsistent alignment!  DONE, thanks.

PR #248, opened 2022-06-30
Move `FunctorApplicativeMonad` to its own namespace, fix `gen-everything script`

mark-moir:
This PR moves `FunctorApplicativeMonad` to its own namespace, so that `Dijkstra.AST.*` is now entirely independent of `Haskell.*`.  Arguably, `FunctorApplicativeMonad` should go away and we should use the definitions in the Agda standard library,  but that's for another day.


mark-moir:
I discussed this with @haroldcarr and am self approving and merging.

